PR #2, opened 2019-04-16
Update README.md

mellogarrett:
- Fix typo

Really cool library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ const App = () => (
 
 ### The power of HOOKS!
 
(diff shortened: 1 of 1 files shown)


molefrog:
Hey @mellogarrett thanks for fixing that! 
Btw, what do you think about the text in README in general? I'm not a native speaker, but it feels like it would benefit from some editing.

malinkie:
I thought the README was a great intro to the library. Thanks for putting this out!

PR #6, opened 2019-04-21
Add `staticHistory` utility for SSR

### LINKED ISSUE #4: Is this SSR compatible ?

molefrog:
Closes #4

This PR introduces a `staticHistory` method (not part of the library by default, should be imported manually), which allows to use the library on the server.

The example usage:

```js
// App.js
import { Router, Route } from "wouter";

const App = (props) => (
  <Router history={props.history}>
    <Route path="/users">...</Route>
  </Router>
);
```

```js 
// server.js
const staticHistory = require("wouter/extra/static-history");

const handleReq = (req, res) => {
  // renders to string
  renderToStaticMarkup(<App history={staticHistory(req.path)} />);
}
```


molefrog:
Still not sure about the `/extra` naming since it kinda conflicts with the library's motto: *nothing extra, just hooks** lol.

Issue #15, opened 2019-04-28
Getting active path


stsourlidakis:
Hello, wouter seems great so far!

I would like to highlight the current page's link on my navbar, is there a better way than the following?

```js
const [, params] = useRoute("/:page");
const currentPage = params ? `/${params.page}` : "/"; //params will be null on /
```

molefrog:
Sure, there is a better way. The `useRoute` hook returns a boolean value, which shows if there was a match. So basically you could extend the current link's behaviour like: https://github.com/molefrog/wouter/pull/17/files

Although I'm still considering if the NavLink should become a part of the package. These types of behaviours are usually quite easy to implement yourself.

molefrog:
Hey @stsourlidakis I have prepared a code sandbox that illustrates a use of `ActiveLink`: https://codesandbox.io/s/5zjpj19yz4. 

I could potentially add this behaviour to the current `Link` component, but this will lead to unnecessary use of context/state in every Link component (unfortunately, you can't place a hook inside a condition). So the only work around is to introduce a new component, but I would rather include it in `extra/` folder.  I'm going to update the README with the list of code recipes and demos.

PR #17, opened 2019-04-29
Add NavLink component

DisaPadla:
Related to [#15](https://github.com/molefrog/wouter/issues/15)

@molefrog  What do you think about this solution?

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,6 +110,9 @@ export const Link = props => {
     event => {
       event.preventDefault();
       navigate(href);
+      if (props.onClick) {
+        props.onClick();
+      }
     },
     [href]
   );
--- a/index.js
+++ b/index.js
@@ -114,7 +114,7 @@ export const Link = props => {
         props.onClick();
       }
     },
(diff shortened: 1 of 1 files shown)


molefrog:
Hey @DisaPadla, I really appreciate the effort, but after thorough consideration I've decided not to include the `NavLink` into the package. The reason is I would love to extend the current `Link` to support function children like this: `<Link>{isActive => ...}</Link>`, but this will lead to extra state and context use in every single link (see this https://github.com/molefrog/wouter/issues/15#issue-438087481)

So ideally it should become a standalone component similar to yours, however I'd like to keep it out of the core library source. My suggestion is following:
1. I'm going to include a new section in the README describing popular code recipes like active links, SSR etc. with code examples
2. Will also probably add the `NavLink` to the `extra/` folder, but the next release.

Again, sorry for rejecting this and thanks for your help!

DisaPadla:
@molefrog  Sure, I agree! Much better add an example to README.

PR #21, opened 2019-05-07
Fix syntax error in Redirect component

StrayFromThePath:


molefrog:
Thanks for that! I'm going to release a new version with a bugfix.

molefrog:
@StrayFromThePath Thanks again for fixing this one! I've just released a new version `v1.2.1` with the bugfix. Two things I also did in order to prevent errors like this from happening:
  -  Added a regression test for `Redirect` component (to make sure it actually changes the current location) https://github.com/molefrog/wouter/commit/3510efe541c02d251aeac35b6bfd77b0dccd06aa
  - Added an ESLint to the project #22, should have done this before 😄 

Great job!

PR #27, opened 2019-05-13
Get rid of `isValidElement` and `Children`

molefrog:
**Why?** because these methods are not provided by Preact and can be replaced with more simplified versions. 

`isValidElement` is replaced with `element && element.props`.
```js
// preact
element && ((element instanceof VNode) || element.$$typeof===REACT_ELEMENT_TYPE)

// react
typeof object === 'object' && object !== null && object.$$typeof === REACT_ELEMENT_TYPE
```

React's `Children` iteration supports things like generators, edge cases like falsey children etc.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,8 +11,7 @@ import {
   createContext,
   isValidElement,
   cloneElement,
-  createElement as h,
-  Children
+  createElement as h
 } from "react";
 
 /*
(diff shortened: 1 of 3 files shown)


molefrog:
After a discussion with @StrayFromThePath we have decided to keep `isValidElement` method and provide an implementation for Preact, since the hack didn't seem to be secure.

Closing this and moving to https://github.com/molefrog/wouter/pull/30

PR #30, opened 2019-05-19
Add support for Preact

### LINKED ISSUE #29: Preact support
Since the library was created to be as small as possible it would make sense to add support for Preact. It should only support the version 10 of Preact, which isn't fully released yet, but supports hooks and fragments. 

There are few interested problems that we will need to solve in order to release this. Here is how I see the final API:

```js
// regular imports
import { Route, Switch, Link } from "wouter/preact";

// additional modules
import staticHistory from "wouter/preact/extra/static-history.js"
```

We can do that by simply creating a new folder `preact` in the repo with one file called `react-deps.js` with all react/preact exports such as `useState`, `createElement` and so on. The folder can be prepopulated with the sources from the root folder on `prepublish` npm script (might need to use this package https://www.npmjs.com/package/copyfiles for cross-platfom support). We shouldn't commit anything except for `preact/react-deps.js` however, the files should be gitginored!

The implementation plan:
  - [x] Remove incompatible methods `isValidElement` and `Children.map`. Merge #27 and check that it doesn't break anything.
  - [x] Create a project structure for Preact `/preact/react-deps.js` and `/react-deps.js` and write a prepublish script.
  - [x] Test the implementation with Preact. Basic rendering, make sure `Route`, `Link` and `Switch` work. A 2-3 test cases should be enough. One thing that isn't clear yet is how to require right `react-deps.js` in tests without having sources in `preact` folder. Maybe Jest mocks?
  - [ ] Describe the Preact feature in the README.

molefrog:
Closes #29 
- [x] Remove `React.Children`
- [x] Create `react-deps.js`
- [x] Write a prepublish script
- [x] Test preact feature


molefrog:
@StrayFromThePath We you're going to test the preact implementation you can mock the `react-deps.js` module like so:

```js
import * as preactDeps from "../preact/react-deps.js";

jest.mock("../react-deps.js", () => preactDeps);

it("now uses preact methods", () => {
  // ...
});
```

react-deps.js
```diff
@@ -0,0 +1,25 @@
...
+  useMemo,
+  useEffect,
+  useState,
+  useContext,
+  useCallback,
+  createContext,
+  isValidElement,
+  cloneElement,
+  createElement
+} from "react";
```
molefrog: Can we just write:

```js
export {
  useRef,
  useMemo,
  useEffect,
  useState,
  useContext,
  useCallback,
  createContext,
  isValidElement,
  cloneElement,
  createElement
} from "react";
```

It's a valid ES6 syntax afaik.

package.json
```diff
@@ -13,16 +14,19 @@
...
     "matcher.js",
-    "extra/"
+    "extra/",
+    "preact/"
   ],
   "scripts": {
     "test": "jest --verbose --coverage",
-    "size": "size-limit"
+    "size": "size-limit",
+    "prepublish": "node scripts/copy.js"
```
molefrog: I would suggest dropping a script file and just inline the command here:
```suggestion
    "prepublish": "copyfiles index.js history.js matcher.js extra/** preact/"
```
Less files is better imo 😎

preact/react-deps.js
```diff
@@ -0,0 +1,27 @@
...
+} from "preact";
+import {
+  useRef,
+  useMemo,
+  useEffect,
+  useState,
+  useContext,
+  useCallback
+} from 'preact/hooks';
+import {isValidElement} from 'preact/compat';
```
molefrog:
```suggestion
export { isValidElement } from 'preact/compat';
```

preact/react-deps.js
```diff
@@ -0,0 +1,27 @@
+import {
+  createContext,
+  cloneElement,
+  createElement
+} from "preact";
+import {
```
molefrog:
```suggestion
export {
```

preact/react-deps.js
```diff
@@ -0,0 +1,27 @@
...
```
molefrog:
```suggestion
export {
```

preact/react-deps.js
```diff
@@ -0,0 +1,27 @@
...
+  useRef,
+  useMemo,
+  useEffect,
+  useState,
+  useContext,
+  useCallback
+} from 'preact/hooks';
+import {isValidElement} from 'preact/compat';
+  
+export {
```
molefrog: Once you do reexport you can drop this part of the file

molefrog:
@StrayFromThePath Also, it appears that `prepublish` is deprecated and is called on install as well. Let's change it to `prepublishOnly`:
Screenshot 2019-05-20 at 19 50 18

molefrog:
Hey @StrayFromThePath I figured why our test wasn't working properly: we should have waited for effects to run before making assertions. Apparently the `preact-test-library` works differently from React one.

Luckily,  I have found a `preact/test-utils` package which is not documented yet https://github.com/developit/preact/blob/10.0.0-beta.1/test-utils/src/index.js but does provide some handy functions for testing. They probably will release it with the version 10.

I also got a feeling that we could collaborate with `preact-test-library` authors and ask if they can:
1. Update the package on npm https://github.com/antsmartian/preact-testing-library/issues/17
2. Provide a less hacky way to wait for updates. https://github.com/antsmartian/preact-testing-library/blob/master/src/index.js#L53

test/preact.test.js
```diff
@@ -1,132 +1,79 @@
 /** @jsx h */
-import { h } from "preact";
+import { h, Fragment, options } from "preact";
```
StrayFromThePath: `options` not used

test/preact.test.js
```diff
@@ -1,132 +1,79 @@
...
-          </Route>
-          <Route path="/users/:name">
-            <h2>Hello, Alex!</h2>
-          </Route>
-          <Route path="/users/:rest*">
-            <h3 />
-          </Route>
-        </Switch>
-      );
+    const { container, getByTestId } = renderResult;
```
StrayFromThePath: `container` not used

PR #41, opened 2019-06-24
add useEffect deps

jeetiss:
🚨


molefrog:
The array of arguments is required, right? Is this documented anywhere?

molefrog:
Oh shit I just found it:
> By default, effects run after every completed render, but you can choose to fire it only when certain values have changed

Which in our case resulted in sub/unsub on every render. Not critical, but wasteful. Thanks for that!

PR #52, opened 2019-07-11
Add `replace` param to `useLocation` updater func

Ty3uK:
Add aibility to call replaceState when using `useLocation` hook:

```js
import { useLocation } from 'wouter';

[, setLocation] = useLocation();

setLocation('/about', true); // calls replaceState instead of pushState
```

I also don't know if I need to add a similar prop to Redirect or Link components?

### DIFF
--- a/test/use-location.test.js
+++ b/test/use-location.test.js
@@ -76,4 +76,15 @@ describe("`update` second parameter", () => {
     expect(history.length).toBe(histBefore + 1);
     unmount();
   });
+
+  it("replaces last entry with a new entry in the History object", () => {
+    const { result, unmount } = renderHook(() => useLocation());
+    const update = result.current[1];
+
+    const histBefore = history.length;
+    act(() => update("/about", true));
+
+    expect(history.length).toBe(histBefore);
+    unmount();
+  });
 });
--- a/test/use-location.test.js
+++ b/test/use-location.test.js
@@ -82,9 +82,10 @@ describe("`update` second parameter", () => {
     const update = result.current[1];
 
     const histBefore = history.length;
-    act(() => update("/about", true));
+    act(() => update("/foo", true));
 
     expect(history.length).toBe(histBefore);
+    expect(location.pathname).toBe("/foo");
     unmount();
   });
 });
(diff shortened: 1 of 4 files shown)


test/use-location.test.js
```diff
@@ -76,4 +76,15 @@ describe("`update` second parameter", () => {
...
   });
+
+  it("replaces last entry with a new entry in the History object", () => {
+    const { result, unmount } = renderHook(() => useLocation());
+    const update = result.current[1];
+
+    const histBefore = history.length;
+    act(() => update("/about", true));
+
+    expect(history.length).toBe(histBefore);
```
molefrog: Can you also add another assertion here just to make sure the location has indeed been changed? 😄

use-location.js
```diff
@@ -13,7 +13,7 @@ export default () => {
     return () => events.map(e => removeEventListener(e, handler));
   }, []);
 
-  return [path, to => history.pushState(0, 0, to)];
+  return [path, (to, replace) => history[replace && "replaceState" || "pushState"](0, 0, to)];
```
molefrog: The expression `replace && "replaceState" || "pushState"` was a bit hard to read. Can we use a ternary operator here maybe? Thanks!

test/use-location.test.js
Ty3uK: Of course, my bad 🙂

molefrog: Not your bad at all, you did everything right here. I just feel the assertion would make us more confident about the code.

use-location.js
Ty3uK: Yes, in the first version I did...

molefrog: Ok, so what made you change that?

Ty3uK: I don't know 🤦‍♂️

test/use-location.test.js
Ty3uK: Done

use-location.js
Ty3uK: Done

molefrog:
@Ty3uK There is also another one here: https://github.com/molefrog/wouter/commit/afe5a4446b9996c50e7ce420e8a67483bd0f0521

Ty3uK:
@molefrog Fixed

molefrog:
Gonna merge this and release the new version once we have types in the repo.

PR #51, opened 2019-07-11
Move typings from DefinitelyTyped

Ty3uK:
According to discussion #49

PR #53, opened 2019-07-11
Add `replace` param to PushCallback type

Ty3uK:
According to #51 and #52.

Also, I think we need to add typings files to `files` field in package.json, right?


molefrog:
@Ty3uK Right, can you do that?

Ty3uK:
@molefrog Of course!

PR #54, opened 2019-07-12
Fix children type in Link interface

StrayFromThePath:
`<Link to="/">Home</Link>`

Link component don't accept text as child element.


molefrog:
See #56

PR #55, opened 2019-07-12
Add module and types settings in package.json for wouter and wouter/preact

cedeber:
These settings are useful for tools like [pikapkg](https://www.pika.dev) which require the module settings in order to build wouter. Adding a private `package.json` into the `/preact` folder allow to do `import [..] from "wouter/preact";`

I don't add the `main` setting because most bundlers are waiting for ES5 here, which wouter doesn't provide. And that's ok ;)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,11 +1,8 @@
 module.exports = function(api) {
   api.cache(true);
 
-  const presets = ["@babel/preset-env"];
-  const plugins = ["@babel/plugin-transform-react-jsx"];
-
   return {
-    presets,
-    plugins
+    presets: ["@babel/preset-env"],
+    plugins: ["@babel/plugin-transform-react-jsx"],
   };
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,5 @@
 module.exports = function(api) {
-api.cache(true);
+  api.cache(true);
 
   const presets = ["@babel/preset-env"];
   const plugins = ["@babel/plugin-transform-react-jsx"];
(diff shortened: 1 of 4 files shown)


cedeber:
The build failure happens because of the `/preact/package.json` file. No idea how to solve this problem.

Ty3uK:
@cedeber To solve that issue move Babel config from `package.json` to `babel.config.js` file.

cedeber:
@Ty3uK Thanks for the help. Looks good now.

molefrog:
@cedeber @Ty3uK Thanks for the help guys. I'm gonna try to review this one Monday.

babel.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function(api) {
+api.cache(true);
```
Ty3uK: Missed indentation?

babel.config.js
```diff
@@ -0,0 +1,11 @@
...
+
+  const presets = ["@babel/preset-env"];
+  const plugins = ["@babel/plugin-transform-react-jsx"];
```
Ty3uK: Why not just write
```js
module.exports = {
  presets: ["@babel/preset-env"],
  plugins: ["@babel/plugin-transform-react-jsx"]
}
```
?

cedeber: This looks new for Babel 7. Apparently setting the cache to true keeps the config in memory for each calls instead of reading it each time. See https://babeljs.io/docs/en/config-files#apicache. Both work. I don't care which one is preferred ;) I simply did it the way it is documented.

Ty3uK: My bad, I mean

```js
return {
  presets: ["@babel/preset-env"],
  plugins: ["@babel/plugin-transform-react-jsx"]
}
```
without unnecessary variables

PR #61, opened 2019-07-19
Extends props interface for component Link

polRk:


polRk:
Or is it better to expand component types with types React.AnchorHTMLAttributes<HTMLAnchorElement> ?

Ty3uK:
@ you can check discussion at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37008

molefrog:
> Or is it better to expand component types with types React.AnchorHTMLAttributes
Yes, that's probably better. Can we somehow tell that the component could accept all attributes from `React.AnchorHTMLAttributes` **and** also wouter-related props?

In the meantime, while types are not fixed yet you could try this workaround, which is actually a preferred way to use link in Wouter (the approach is similar to [using links in Next.js](https://nextjs.org/docs#with-link))
```js
<Link to="/app">
  <a className="link__container"> ... </a>
</Link>
```

molefrog:
Hey @polRk, take a look at this PR #64
I've added a few tweaks to our typings and added type specs to better ensure that they indeed work. If this looks ok, we can merge the PR and release the new version with types bundled.

PR #62, opened 2019-07-22
Type specs (React only)

### LINKED ISSUE #56: Minor type tweaks to cover edge cases
Hey @StrayFromThePath @StrayFromThePath I'm super excited to have types finally in the project! I managed to spent some time reading the sources and although I'm not a TS user I've got some feedback I thought I need to share with you. 

### Proper `Params` type
The current `Params` type is a union of hash or `null`. This works fine for the `Match` type which can represent either `[true, { ... }]` or `[false, null]`, however this also assumes that the render prop can contain the null as a first argument [here](https://github.com/molefrog/wouter/blob/master/index.d.ts#L22). I think we can narrow the type even further, because the render prop always guarantees that the argument is a valid object. 

So how can this be solved? I suggest we make the Params type to always be an object and mark that the match result could contain null:
```js
type Params = { [paramName: string]: string }

// it this a real TypeScript code? :) I'm not sure, but you get the idea
type Match = [boolean, Params | null];

// or is this even possible:
type Match = [true, Params] | [false, null];
```

### Remove `match` props from `Route`
This prop is an internal thing used by the switch, so it's not part of the public API. I suggest we either remove it or change the type, since it's not a boolean anymore, but `Match`. But I think it's easier to just remove this. 

### Proper type of `children` in `Router`
As @StrayFromThePath pointed out in #54, some components need proper types to support mixed React nodes. The current type of the `children` prop in `Router` is `ReactElement | ReactElement[];`, but this makes it impossible to use something like:

```js
<Router>
  Hello, this is an <Application />
  We also have <Footer /> and some {1337} numbers here.
</Router>
```
 
Same applies to `Link`. Generally speaking, anything that `React.Provider` accepts should work here as well. 

Let me know what you guys think and keep up with the good work!

molefrog:
Closes #56

- [x] Move React types into types/ folder. Things like `tsconfig.json` create to much overhead in the root folder, it looks way better when everything TS-related lives in its own folder IMO.
- [x] Make sure types are published correctly 
- [x] Write type specs
- [x] Remove DT banner
- [x] Integrate with CI 

I had to remove the banner from `index.d.ts` since these types do not live inside DefinitelyTyped and `dtslint` throws a warning:
> Error: Since this type definition includes a header (a comment starting with `// Type definitions for`), assumed this was a DefinitelyTyped package.
But it is not in a `DefinitelyTyped/types/xxx` directory: /Users/molefrog/git/wouter/types


Ty3uK:
@molefrog fixed copule mistakes at #63

PR #64, opened 2019-07-23
Types: allow `<Link />` component to accept anchor element attrs

molefrog:
Makes it possible to use any prop that a normal link accepts.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,6 +3,7 @@
 // * Maksim Karelov <https://github.com/Ty3uK>
 
 import {
+  AnchorHTMLAttributes,
   FunctionComponent,
   ComponentType,
   ReactElement,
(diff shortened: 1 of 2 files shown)


polRk:
Why do we make these fields optional
```
  to?: Path;
  href?: Path;
```

And why do we use these fields
```
  to?: Path;
  href?: Path;
  onClick?: () => void;
  children: ReactNode;
```

Maybe it's better
```
  to: Path;
  children: ReactNode;
```

molefrog:
@polRk `href` and `to` are essentially aliases and are mutually exclusive: either one of them should be present, but not both at the same time. This was introduced for better DX. We weren't sure how to describe this in TS, maybe some algrebraic type operations?

As for the `onClick` prop — I agree it should be part of anchor attributes, the problem is — our current handler doesn't provide an argument, while in fact it should. I can fix that in another PR, which will also include some source code changes.

PR #66, opened 2019-07-24
Tweak Link component types

### LINKED ISSUE #65: Pass `event` to onClick handler
This will allow to use proper types.

molefrog:
Closes #65
  - `href` and `to` are mutually exclusive
  - `onClick` now accepts an event argument and can be compatible with `onClick` from `AnchorHTMLAttributes`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -36,8 +36,6 @@ export const Route: FunctionComponent<RouteProps>;
 export interface LinkProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
   to?: Path;
   href?: Path;
-  onClick?: () => void;
-  children: ReactNode;
 }
 export const Link: FunctionComponent<LinkProps>;
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -33,16 +33,17 @@ export interface RouteProps {
 }
 export const Route: FunctionComponent<RouteProps>;
 
-export interface LinkProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
-  to?: Path;
-  href?: Path;
-}
+export type NavigationalProps =
+  | { to: Path; href?: never }
+  | { href: Path; to?: never };
+
+export type LinkProps = Omit<AnchorHTMLAttributes<HTMLAnchorElement>, "href"> &
+  NavigationalProps;
+
 export const Link: FunctionComponent<LinkProps>;
 
-export interface RedirectProps {
-  to?: Path;
-  href?: Path;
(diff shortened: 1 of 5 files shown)


molefrog:
Hey @StrayFromThePath @polRk @Ty3uK, check this out. I've finally managed to construct a proper type that guarantees that `href` and `to` are mutually exclusive: either one of them is used, but not both.

types/index.d.ts
```diff
@@ -33,16 +33,17 @@ export interface RouteProps {
 }
 export const Route: FunctionComponent<RouteProps>;
 
-export interface LinkProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
-  to?: Path;
-  href?: Path;
-}
+export type NavigationalProps =
```
polRk:
```typescript
export type NavigationalProps = { to: Path } | { href: Path}
```

types/type-specs.tsx
```diff
@@ -51,8 +52,14 @@ const invalidParams: Params = { id: 13 }; // $ExpectError
 /*
  * Link and Redirect component type specs
  */
+
+// `to` and `href` are aliases, but they are mutually exclusive and
+// can't be used at the same time:
 <Link to="/users">Users</Link>;
 <Link href="/about">About</Link>;
+<Link href="/about" to="/app" children="" />; // $ExpectError
```
polRk: delete

types/type-specs.tsx
```diff
@@ -51,8 +52,14 @@ const invalidParams: Params = { id: 13 }; // $ExpectError
...
+<Link children="" />; // $ExpectError
```
polRk: delete

types/index.d.ts
polRk: It needs to be tested, I'm sure that it works, but not 100%

types/type-specs.tsx
molefrog: Why should we delete those? These lines assert the error: when both `href` and `to` are given TypeScript will throw an error. The `// $ExpectError` is a magic comment that tells `dtslint` that this lines **should raise** a type error.

types/type-specs.tsx
molefrog: Same here. By specifying `// $ExpectError` we tell linter that this line is incorrect in terms of types: either `to` or `href` should be present.

types/index.d.ts
molefrog: Well, this is tested via the `Link` component tests. See comments below.

PR #77, opened 2019-08-19
Use effect should only ever return a function

timmak:
Currently use effect return the result of `push()` return the length of the array this will cause an warning inside of react.

### DIFF
--- a/index.js
+++ b/index.js
@@ -146,8 +146,8 @@ export const Switch = ({ children, location }) => {
 export const Redirect = props => {
   const [, push] = useLocation();
   useEffect(() => {
-    push(props.href || props.to)
-  });
+    push(props.href || props.to);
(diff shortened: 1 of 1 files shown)


molefrog:
@timmak Good one, thanks! We should also probably add an empty array as a second argument to make sure redirect is only fired on component mount. Can you take care of that?

Also, can you run Prettier over modified files, think there is a semicolon missing.

timmak:
Should be good now

molefrog:
@timmak How soon do you need this to be published? I'm planning to implement a few other features before creating a release.

timmak:
Can wait will just point to commit for now, as it not going to production anytime soon.

molefrog:
@timmak Noted! Thanks for the PR.

PR #80, opened 2019-08-21
Support a basepath config in <Router/>

cbbfcd:
I really like this library, I want to get involved and make this library a bit cooler.
I noticed a discussion about `basepath`, my plan is like this:

```javascript
<Router basepath='/app'>
  <Route path='/about' />
</Router>

// Link also have this ability
<Link href='/about' />

// also
<Link href='/about'>
 { props => <span><a href={props.href}>click here</></span>}
</Link>
```
then, when the link be clicked, will navigate to the url `/app/about`。
so, It will be no burden for the user.

Of course, I can make it more detailed, but I think it is enough now.

Looking forward to your review!

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,8 @@ const RouterCtx = createContext({});
 const buildRouter = (options = {}) => {
   return {
     hook: options.hook || locationHook,
-    matcher: options.matcher || makeMatcher()
+    matcher: options.matcher || makeMatcher(),
+    basepath: options.basepath || ''
   };
 };
 
@@ -43,10 +44,10 @@ export const useLocation = () => {
 };
 
 export const useRoute = pattern => {
-  const router = useRouter();
+  const { matcher, basepath } = useRouter();
   const [path] = useLocation();
 
-  return router.matcher(pattern, path);
+  return matcher(basepath + pattern, path);
 };
 
 /*
@@ -83,9 +84,10 @@ export const Route = ({ path, match, component, children }) => {
 };
 
 export const Link = props => {
+  const { basepath } = useRouter();
   const [, navigate] = useLocation();
 
-  const href = props.href || props.to;
+  const href = basepath + (props.href || props.to);
   const { children, onClick } = props;
 
   const handleClick = useCallback(
@@ -110,6 +112,9 @@ export const Link = props => {
 
   // wraps children in `a` if needed
   const extraProps = { href, onClick: handleClick, to: null };
+
+  // render props
+  if (typeof children === 'function') return children(props);
   const jsx = isValidElement(children) ? children : h("a", props);
 
   return cloneElement(jsx, extraProps);
--- a/index.js
+++ b/index.js
@@ -77,17 +77,16 @@ export const Route = ({ path, match, component, children }) => {
   if (!matches) return null;
 
   // React-Router style `component` prop
-  if (component) return h(component, { params: params });
+  if (component) return h(component, { params });
 
   // support render prop or plain children
   return typeof children === "function" ? children(params) : children;
 };
 
 export const Link = props => {
-  const { basepath } = useRouter();
   const [, navigate] = useLocation();
 
-  const href = basepath + (props.href || props.to);
+  const href = props.href || props.to;
   const { children, onClick } = props;
 
   const handleClick = useCallback(
@@ -104,7 +103,7 @@ export const Link = props => {
         return;
 
       event.preventDefault();
-      navigate(href);
+      navigate(href, false, props.state);
       onClick && onClick(event);
     },
     [href, onClick, navigate]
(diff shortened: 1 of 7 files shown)


molefrog:
@cbbfcd Hi! Thanks for the PR, I've got some feedback, but I'm going to need a few days to prepare it here.

cbbfcd:
@molefrog  Looking forward to your recovery, in addition, I will continue to think about the search, state, and related discussions in the Issue

molefrog:
Hi @cbbfcd and thanks for waiting. I've finally found time to check the PR.

I really like the idea with configuring the basepath through the router! One thing I noticed is with current implementation we will also need to properly path `Switch` and `Redirect` to make sure all parts of router support basepath (e.g. when the redirect is given `<Redirect to="/users" />` the actual location should be `/app/users` with the basepath. We could potentially path all these components but ... I have a better idea!

The core of the router is `useLocation` hook, so if we could make it possible for it to handle basepath transparently, all other components will work automatically. In fact, that's way more secure. What I mean is something:

```js
const [location, setLocation] = useLocation({ basepath: "/app" });

// location is /users
// when setLocation is called with /orders the actual url is /app/orders
```
And we will also need to inject the logic into Link of course, since it should generate proper absolute urls:
```js
<Link href="/account" />
// actually renders <a href="/app/account" /> but 
// continues to call setLocation("/account") internally
```
So Link is the only component that will know about the basepath and I think that's totally ok!
So I see two options here:
  - Either path the `useLocation` hook here to properly handle basepath option and make all needed transformations over the result returned by the `router.hook()`
  - Or, which is even cooler: make it possible in `use-location.js` default hook implementation to accept a `{ basepath: "/app" }` as an argument. In fact, the object is already there, since hook is called with router as a first argument: `router.hook(router) // router is just an object`.

I personally like the last solution since:
1. Adds a basepath functionality to the default hook itself, which can be used as a standalone hook (https://github.com/molefrog/wouter#1kb-is-too-much-i-cant-afford-it)
2. Allows us to save some bytes, since router is already present as a first argument of the hook.

Looking forward to your feedback.

cbbfcd:
@molefrog 
Thank you very much for receiving your reply, I very much agree with your thoughts.
`use-location.js` is the place that should be empowered.
replacing greater power with minimal changes and guaranteeing the independence of `useLocation`. it's so cool!!
but, Can we do more?

like this:

```javascript
export default ({basepath = '', ...others}) => {

  // getLocation will return an obj which looks like {pathname, search, state}
  const getLocation = () => ({
    pathname: basepath + location.pathname,
    search: location.search,
    state: history.state
  });
 // not only the path any more
  const [loc, update] = useState(getLocation());

  useEffect(() => {
    patchHistoryEvents();

    const events = ["popstate", "pushState", "replaceState"];
    // there should be a magic handle here to avoid re-render
    const handler = () => update(getLocation());

    events.map(e => addEventListener(e, handler));
    return () => events.map(e => removeEventListener(e, handler));
  }, []);

  // the 2nd argument of the `useLocation` return value is a function
  // that allows to perform a navigation.
  //
  // the function reference should stay the same between re-renders, so that
  // it can be passed down as an element prop without any performance concerns.
  const navigate = useCallback(
     // wo can navigate with a param
    (to, replace, state = {}) => history[replace ? "replaceState" : "pushState"](state, 0, to),
    []
  );

  return [loc, navigate];
};
```
This is just a draft and I look forward to your reply.

cbbfcd:
@molefrog 
I practiced some ideas：
1. `navigate`: 
we can add basepath here and pass the parameter through [history.state](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method)
```js
const navigate = useCallback(
    (to, replace, state) => history[replace ? "replaceState" : "pushState"](state, 0, basepath + to),
    []
);
```
the problem here is that there is no history and location in the` ssr ` scenes. For this, I think the [reach/router scheme](https://github.com/reach/router/blob/master/src/lib/history.js#L69) is simple enough and effective.
I haven't found a good solution yet.
2. i modified the rules that match the default route.
```js
//before
<Switch>
  <Route path="/about">...</Route>
  <Route path="/:rest*">404, not found!</Route>
  <Route path="/">...</Route> // miss....
</Switch>

// now
<Switch>
  <Route path="/about">...</Route>
  <Route default>404, not found!</Route>
</Switch>
```
this way I can avoid managing the placement of the default route. and with basepath the match will cause trouble: if basepath is `/app`, then `/app/:anything*` will not be matched with '/'.
3. i add some unit test cases which prove that my modifications are valid.

molefrog:
Hey @cbbfcd please take a look at #94 which is based on your ideas, but uses a bit different approach: the basepath logic is completely handled insead a `useLocation` method. 

@omgovich and I have been looking at your PR and first wanted to add some commits to it, but then decided to create a new one from scratch. Hope you don't mind! Please do leave code comments and a code review is welcome!

cbbfcd:
@molefrog Great, thank you for your reply!

cbbfcd:
I will continue to pay attention, and will discuss some ideas for expansion in the issue.

PR #85, opened 2019-09-26
Delay redirecting until "history" was hooked in Switch

3846masa:
"Switch" cannot detect redirect immediately after mounting.
Sample code: https://codesandbox.io/s/flamboyant-merkle-zqm4c?fontsize=14

```jsx
<Switch>
    <Route path="/hello">Hello Wouter!</Route>
    <Route path="/:rest*">
      If you see this, Wouter cannot detect the redirect.
      <Redirect to="/hello" />
    </Route>
</Switch>
```

The function called from "useEffect" is called first from the function of the child element.
Because of that, "Switch" cannot hooks history before calling `useEffect` in "Redirect" for redirecting.

This PR fixes that issue.
The solution is, "Redirect" calls `useEffect` twice to delay redirecting.
At first `useEffect` calling, "Redirect" does nothing, but "Switch" hooks history.
At 2nd `useEffect` calling, "Redirect" redirects, and emit event.
"Switch" already hooks history, so "Switch" can detect redirect immediately after mounting.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ import makeMatcher from "./matcher.js";
 
 import {
   useRef,
+  useState,
   useEffect,
   useContext,
   useCallback,
@@ -140,13 +141,18 @@ export const Switch = ({ children, location }) => {
 };
 
 export const Redirect = props => {
+  const [delayed, update] = useState(false);
   const [, push] = useLocation();
   useEffect(() => {
-    push(props.href || props.to);
+    if (delayed) {
+      push(props.href || props.to);
+    } else {
+      update(true);
+    }
 
-    // we pass an empty array of dependecies to ensure that
+    // we pass only `delayed` as dependecies to ensure that
     // we only run the effect once after initial render
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [delayed]); // eslint-disable-line react-hooks/exhaustive-deps
 
   return null;
 };


molefrog:
Hey @3846masa ! 
Thanks for the contribution and sorry for the long response. Was super overwhelmed these couple of months. I'll need some time to research the behavior you reported, it looks a bit tricky indeed. 

Will get back to you as soon as possible.

molefrog:
Ok, so I've investigated the problem and it seems like the problem is in `useLocation` hook. It looks like the location listener for the Switch is actually added after the redirect happens:
Screenshot 2019-10-08 at 16 18 14

It could be related to the fact the we use `useEffect` to set up an event listener. If I change it to `useLayoutEffect` it starts to work properly. I might need some time to check that it's safe to migrate to `useLayoutEffect`.

molefrog:
@3846masa Ok, so after some thinking I was able to adapt @bvaughn's snippet https://gist.github.com/bvaughn/e25397f70e8c65b0ae0d7c90b731b189 to properly catch the updates that might happen between the first render and `useEffect` call. You can see the patch here:

https://github.com/molefrog/wouter/commit/f8ea4ee021fd4f31af14d1ea794645fa8fd84c7e

I've also published it as a part of the new version `v2.3.0`, which should fix the problem you have experienced. @3846masa Can you verify if it works now?

3846masa:
I tried v2.3.0, it seems to work fine!
https://codesandbox.io/s/cocky-antonelli-3elft?fontsize=14

Thanks, @molefrog !!

molefrog:
Nice!

PR #87, opened 2019-10-07
TS types: add generic for Params

ArnaudBarre:
Thanks for this lib and bundling the types !

React hooks types use a lot generics and I would like to get the same with wouter.

Sadly I ran into the [no-unnecessary-generics](https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md) of dtslint. The doc suggest that I should use the `as` keyword but even [@types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L949) provide generics. There is a [stale issue](https://github.com/microsoft/dtslint/issues/76) about this. 

If you agree I will make the same changes for preact types

Note: I added yarn.lock for convenience, but I can remove it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 npm-debug.log*
 node_modules/
 .npm
+yarn.lock
 
 # ignore package-lock.json since it's a package
 # see https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514
--- a/.gitignore
+++ b/.gitignore
@@ -2,11 +2,11 @@
 npm-debug.log*
 node_modules/
 .npm
-yarn.lock
 
 # ignore package-lock.json since it's a package
 # see https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514
 /package-lock.json
+/yarn.lock
 
 # bundler
 dist/
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -16,22 +16,24 @@ export type PushCallback = (to: Path, replace?: boolean) => void;
 export type LocationTuple = [Path, PushCallback];
 export type LocationHook = () => LocationTuple;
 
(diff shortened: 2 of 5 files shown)


molefrog:
@ArnaudBarre Hey, thanks a lot for the PR, it looks amazing! I'm not a TS user myself, so wasn't sure how to properly add types for params. But the change makes a lot of sense!

Do I get it right that the current projects using wouter's types will continue to work, since you specify a default generic parameter?

> Note: I added yarn.lock for convenience, but I can remove it.
Don't mind having it in gitignore. Can you just maybe move it below the `package-lock.json` line?

P.S. Would also love to hear in what kind of projects you're using wouter? 😄

ArnaudBarre:
> Do I get it right that the current projects using wouter's types will continue to work, since you specify a default generic parameter?

Yes exactly. A potential breaking change is if people where expecting non nullable props other thats params for a route component. But it's unlikely and it will happen at compile time and probably prevent a bug

> P.S. Would also love to hear in what kind of projects you're using wouter?

For some small projects in my company and in the main product which is under development (patient-doctor communication tool for chronic disease)

molefrog:
> For some small projects in my company and in the main product which is under development (patient-doctor communication tool for chronic disease)

Awesome! Please do let me know when the project is live, so I could add it to the list of the companies using wouter. If you don't mind of course!

I'm going to merge this PR, it looks absolutely ok to me. I'm planning to release `v2.2.2` with the change and maybe some other improvements I'm currently working on.

ArnaudBarre:
Not anytime soon but I will do ! Thanks for the quick responses !

molefrog:
@ArnaudBarre Released it in `v2.3.0`

PR #88, opened 2019-10-09
Use isValidElement from preact core

developit:
It's been merged into core, in part because we want Wouter to have first-class support!

/cc @kristoferbaxter

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "eslint-plugin-react-hooks": "^1.7.0",
     "jest": "^24.9.0",
     "jest-esm-jsx-transform": "^1.0.0",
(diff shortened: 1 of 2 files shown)


molefrog:
@developit Awesome! Going to merge it right away. 

By the way, we now ship a separate Preact-only package `wouter-preact` which now contains a proper list of peer deps: https://github.com/molefrog/wouter/blob/master/preact/package.json

As far as I understand this change isn't compatible with all release candidates, so I'll also set the Preact version to `^10.0.0` there.

molefrog:
Alright, so I've just published `wouter@2.3.0` and `wouter-preact@2.3.0` with full Preact X support. Thanks for the help @developit !

PR #92, opened 2019-10-26
Allow passing replace & state props

### LINKED ISSUE #75: Allow passing state to link/setLocation
Would it be possible to support the following using the browser's `history.state` API, optionally allowing `history[method](stateObj, 0, to)`, instead of always using `history.[method](0, 0, to)`?

e..g.

```jsx
<Link href="/" state={{ exampleState: true }}>Link</Link>
```
and
```js
const [location, setLocation] = useLocation();

setLocation("/", { exampleState: true });
```

Not sure where the accessor makes the most sense, maybe something like:
```js
import { useRouterState } from 'wouter';

const routerState = useRouterState();
```

Thanks!

soulchainer:
Closes #75

Allow passing `replace` & `state` props to `Link`, `Redirect` and `useLocation`.

`useLocation` already has the possibility to pass `replace`, but that wasn't being used by `Link` and `Redirect` (which should be allowed, as far as I remember, most of the routing libraries with that feature are doing it). I had to add it because that (consistency), plus anyways I needed to pass the third value for the `state`.

Didn't create an extra accessor for the history `state` because I felt it kinda logic this way. I mean, when we add a `state`, we want it to use in the location we're moving to at that moment, so it sounds logic to me to add it there.
I edited it considering to not make any breaking change and that prior code worked without changes. I hope I didn't miss anything.

Also updated the `README.md` with the changes and the tests accordingly. 

I have been monitoring #75 a few months now, waiting for some action or maybe to @destructobeam to make a PR, and like I see no movement, I just make a try. If you feel this wrong o whatever just discard this and have my apologies :).

In any case, I wanted to use `wouter` this summer to migrate a tiny silly project of mine (used just for experimenting), but like this feature (`state`) wasn't supported, I had to discard it. For me, this is a key feature and IDK, most route libraries support it because native `history` also support it, so.

I tried to do the fewer changes possible, but anyways `yarn run size` fails, because it looks like... the size limit is just adjusted to the current size, right? So, if this looks good (or after some revisions, it ends looking that way), that limit should be upped in `package.json` and `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,12 +96,15 @@ a navigation with it, this is very similar to how you work with values returned 
 import { useLocation } from "wouter";
 
 const CurrentLocation = () => {
-  const [location, setLocation] = useLocation();
+  const [location, setLocation, state] = useLocation();
 
   return (
     <div>
       {`The current page is: ${location}`}
+      {`The current page location state is: ${state}`}
       <a onClick={() => setLocation("/somewhere")}>Click to update</a>
+      <a onClick={() => setLocation("/somewhere", true)}>Click to update replacing the last entry in the browsing history</a>
+      <a onClick={() => setLocation("/somewhere", false, { from: '/here' })}>Click to update with an additional location state object</a>
     </div>
   );
 };
@@ -193,7 +196,7 @@ import { Route } from "wouter";
 <Route path="/orders/:status" component={Orders} />
 ```
 
-### `<Link href={path} />`
+### `<Link href={path} replace state={stateObj} />`
 
 Link component renders an `<a />` element that, when clicked, performs a navigation. You can customize the link appearance
 by providing your own component or link element as `children`:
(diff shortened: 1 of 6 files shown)


destructobeam:
I haven't been doing much work on the (personal) project I am using this for due to client work, but this looks great to me!

molefrog:
Hello and thanks for the wonderful PR! 

I would love to merge your PR right away, but there are a few things I'd like to improve and address in terms of the design of this feature. It's a bit challenging since I have a few ideas for the next release such as: query params, basepath config etc. And they are a bit tightly coupled. 

What I would love to have is to express my ideas in a separate issue and invite you guys to check it out before we actually roll out any code. Sorry for the bureaucracy, it's just wouter is tiny and every design and API decision should be carefully considered.  

I don't want to look selfish and irresponsible, but I'm a bit overwhelmed with tasks at my day job and don't think I will be able to provide anything this week. Hope for your undestanding!

destructobeam:
@molefrog No worries on my end, I know what it's like to feel strongly about a project.

I would however suggest maybe putting down some feelings around how contributions should be made (if at all) in a CONTRIBUTING.md file in the repo? The project is gettting popular enough that defining some basic process might be worthwhile.

Otherwise looking forward to the discussion issue. Cheers!

soulchainer:
@molefrog OK, I understand. Don't worry, man :D.
Due to the size constraints, I guess it's almost unavoidable to have some logic a bit coupled >_<.

And nothing good comes out from rushing things, so take your time ;).

BTW, I agree with @molefrog on the `CONTRIBUTING.md` file stuff. Seems like the project could use it.

molefrog:
@soulchainer Thanks for waiting! Please join our discussion here where we discuss the design of this feature: #102

molefrog:
With the latest wouter release it's now possible to write location hooks that can customize the navigation. I've written some of the ideas here: https://github.com/molefrog/wouter/issues/102#issuecomment-654694679

I hate to do that, but I have to close the PR, as the state should probably be implemented as an external location hook library. Sorry for taking so long to finally reject the changes. P.S. If you'd like to participate in the development of the location-with-state library, I'd be happy to review and merge your changes! Thanks again for dedicating your time to the library.

PR #97, opened 2019-11-18
update type definition

cbbfcd:
update type definition, because we have `basepath` now!

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -14,7 +14,7 @@ export type Path = string;
 export type PushCallback = (to: Path, replace?: boolean) => void;
 
 export type LocationTuple = [Path, PushCallback];
-export type LocationHook = () => LocationTuple;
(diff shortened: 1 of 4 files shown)


types/index.d.ts
```diff
@@ -62,6 +62,7 @@ export const Switch: FunctionComponent<SwitchProps>;
 
 export interface RouterProps {
   hook: LocationHook;
+  basepath: string;
```
molefrog: While it's not really critical, but I would suggest using a `Path` type here (which is just an alias to `string`, but helps to keep the code more generic I guess). Basically `Path` represents something that can describe a location.

omgovich:
@cbbfcd thank you very much! You help us a lot

types/index.d.ts
cbbfcd: good idea!! Thx! i will fix it

PR #98, opened 2019-11-18
Reduce `useRoute` size

omgovich:
Reduces index.js size by 3 bytes


cbbfcd:
@omgovich  😄, I also think that reducing the code size is a necessary task.

PR #99, opened 2019-11-18
reduce unnecessary calculations

cbbfcd:
Reduced unnecessary calculations and reduced code size.

### DIFF
--- a/index.js
+++ b/index.js
@@ -67,10 +67,8 @@ export const Router = props => {
 };
 
 export const Route = ({ path, match, component, children }) => {
-  const useRouteMatch = useRoute(path);
-
   // `props.match` is present - Route is controlled by the Switch
-  const [matches, params] = match || useRouteMatch;
+  const [matches, params] = match || useRoute(path);
 
   if (!matches) return null;
 


index.js
```diff
@@ -67,10 +67,8 @@ export const Router = props => {
 };
 
 export const Route = ({ path, match, component, children }) => {
-  const useRouteMatch = useRoute(path);
-
   // `props.match` is present - Route is controlled by the Switch
-  const [matches, params] = match || useRouteMatch;
+  const [matches, params] = match || useRoute(path);
```
molefrog: I'm afraid this won't work because of the way hooks are implemented. Hooks are using an internal stack of the calls, so their order matters. One of the drawbacks is that we can't call they inside a condition. More info here: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

> Don’t call Hooks inside loops, conditions, or nested functions. Instead, always use Hooks at the top level of your React function. By following this rule, you ensure that Hooks are called in the same order each time a component renders. That’s what allows React to correctly preserve the state of Hooks between multiple useState and useEffect calls.

molefrog: While the real case when this breaks is difficult to reproduce (we need to render a Route that is first controlled via a `match` prop and then rerenders with `match: undefined`), it's still unsafe because we leave it on a user. And I won't sure this will work with concurrent mode.

cbbfcd: I know this detail, focusing on the reduction in code size, so I ignore this, you are right.

PR #100, opened 2019-11-19
Rename `basepath` to `base`

### LINKED ISSUE #76: Incorrect Link href with basepath
Hello, I'm trying to implement basepath into my application. I followed the README's [instruction](https://github.com/molefrog/wouter#customizing-the-location-hook). Everything works well, except **`<Link />`** component.

For example,

``` jsx
<Link href="/path">link</Link>
```
It's fine that clicking on it will navigate to `:basepath/path`, but the href prop is still `/path`.

Now I had to use `<Link />` in this way,

``` jsx
<Link href="/path">
  <span><a href={`${basepath}/path`}>link</a></span>
</Link>
```

Any idea on it?

omgovich:
Closes #76 

Reduce the bundle size by 5 bytes.

@cbbfcd could you check out the types?

### DIFF
--- a/types/type-specs.tsx
+++ b/types/type-specs.tsx
@@ -167,7 +167,7 @@ if (params) {
   params; // $ExpectType null
 }
 
-const [] = useLocation({base: '/app'}); // $ExpectError
+const [] = useLocation({basename: '/app'}); // $ExpectError
 const [, setLoc] = useLocation({base: '/base'});
 
 setLoc('/app');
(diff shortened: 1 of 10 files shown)


types/type-specs.tsx
```diff
@@ -168,7 +168,7 @@ if (params) {
 }
 
 const [] = useLocation({base: '/app'}); // $ExpectError
```
molefrog: Can you also fix this line as well? It was expecting an error on unknown options, but since `base` now is a valid name we could rename it to:
```suggestion
const [] = useLocation({foo: '/app'}); // $ExpectError
```

cbbfcd:
@omgovich Sorry, a busy day! But you have already solved it~ 👍

Issue #101, opened 2019-11-19
Discussion about default routes


cbbfcd:
i wrote a [demo](https://codesandbox.io/s/wouter-catch-all-route-3g2fu):

```js
<Switch>
    <Route path="/:anything*">Default Route: nothing found!</Route>
    <Route path="/users/one">Users One</Route>
    <Route path="/">First Route</Route>
     //...
    {/* <Route path="/:anything*">Default Route: nothing found!</Route> */}
</Switch>
```
If we accidentally put the default route in front, it will cause problems.

I think there are two ways:
1. The problem is explained in the documentation, which draws the attention of developers⚠️
2. If you use `<Route default component={AnyComponent}/>` and the position is irrelevant, is it a good idea?

@omgovich @molefrog Looking forward to your reply.

omgovich:
Hi @cbbfcd. I just checked out the react-router and found out that they use the same approach where order matters. So we probably shouldn't change anything in code.

molefrog:
Hey guys! I think we should definitely implement support for `default` prop since a lot of people were asking about that and not everyone gets the `:rest*` trick — it's just not developer-friendly. 

Should the order matter is a good question. Rather than copying it 1-1 from react-router I would first think about possible design choices we would need to make in both cases. 

Let's say the order doesn't matter. Can the Switch then have multiple default routes? I think it can, it's valid, although not a very practical scenario. In this case which one gets rendered: first one, second or both? 

On the other hand, we can just say that default Route is a Route that always matches in a Switch. In this case this:

```js
<Switch>
  <Route default>foo!</Route>
  <Route path="/app">app</Route>
  <Route default>Second default!</Route>
</Switch>
```
Would always render `foo!` no matter what (even if the location is `/app`). And it fits better with the Switch contract "only one child is renderer at most at the time".

molefrog:
To recap.
**Position-independent default Route:**
✔️Possible better DX, although breaks the common "catch-all" mental model from server-side routing
❓Isn't clear how it should behave in edge cases
❓Probably harder to implement and is size-consuming: will need to make an extra iteration over children to determine the first default route

**Position-dependent default Route:**
✔️Follows the same mental model used in RR and server-side routing (Rails, Express, Sinatra etc.)
✔️Doesn't break the Switch contract: 
✔️Should be relatively easy to implement  — add an extra check [here](https://github.com/molefrog/wouter/blob/master/index.js#L132):
```diff
[...]

PR #103, opened 2019-11-22
Add another default route matching rule

cbbfcd:
[#101 ](https://github.com/molefrog/wouter/issues/101), we hava a size trouble, +5b～

### DIFF
--- a/test/switch.test.js
+++ b/test/switch.test.js
@@ -59,22 +59,21 @@ it("ignores mixed children", () => {
   expect(rendered[0].type).toBe(Route);
 });
 
-it("ignores other elements", () => {
+it("matches regular components as well", () => {
   const Dummy = props => props.children;
 
   const result = testRouteRender(
     "/",
     <Switch>
+      <Dummy path="/">Component</Dummy>
       <b>Bold</b>
-      <Dummy>Component</Dummy>
-      <Route path="/">route</Route>
     </Switch>
   );
 
   const rendered = result.children[0].children;
 
   expect(rendered.length).toBe(1);
-  expect(rendered[0].type).toBe('b');
+  expect(rendered[0].type).toBe(Dummy);
 });
 
 it("allows to specify which routes to render via `location` prop", () => {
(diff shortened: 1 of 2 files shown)


test/switch.test.js
```diff
@@ -74,7 +74,7 @@ it("ignores other elements", () => {
   const rendered = result.children[0].children;
 
   expect(rendered.length).toBe(1);
-  expect(rendered[0].type).toBe(Route);
+  expect(rendered[0].type).toBe('b');
```
molefrog: What about this? Looks like it changes the behaviour for non-Route elements/components. Should we just ignore this or keep the backward compatibility? 🤔

cbbfcd: I think the current behavior is in line with expectations, it should be the behavior that developers need to control, and I observe that react-router also maintains consistent behavior.

molefrog: Ok, can we replace the test case with:
```js

it("matches regular components as well", () => {
  const Dummy = props => props.children;

  const result = testRouteRender(
    "/",
    <Switch>
      <Dummy path="/">Component</Dummy>
      <b>Bold</b>
    </Switch>
  );

  const rendered = result.children[0].children;

  expect(rendered.length).toBe(1);
  expect(rendered[0].type).toBe(Dummy);
});
```

test/switch.test.js
```diff
@@ -113,3 +113,41 @@ it("always ensures the consistency of inner routes rendering", async () => {
 
   unmount();
 });
+
+it("default route via position-dependent with wildcard expression", async () => {
```
molefrog:
```suggestion
it("supports catch-all routes with wildcard segments", async () => {
```

molefrog: Just to make it more aligned with the "it ..." phrase.

test/switch.test.js
```diff
@@ -113,3 +113,41 @@ it("always ensures the consistency of inner routes rendering", async () => {
...
+    </Switch>
+  );
+
+  const rendered = result.children[0].children;
+
+  expect(rendered.length).toBe(1);
+  expect(result.findByType("h2")).toBeTruthy();
+});
+
+it("default route via position-dependent and without a path prop", async () => {
```
molefrog:
```suggestion
it("uses a route without a path prop as a fallback", async () => {
```

cbbfcd:
👌

molefrog:
@cbbfcd Nice job! I'm going to merge this and we will have this feature in the next release!

cbbfcd:
@molefrog Thanks for your CR, Help me a lot. 😄

PR #107, opened 2019-11-28
Remove not used Route imported component

midudev:
Route is not being used on the example so it's safe to remove from the import.


molefrog:
😉 Nice catch!

PR #122, opened 2020-07-08
 Small optimization, less code size

cbbfcd:
😄

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,7 +60,7 @@ const useNavigate = (options) => {
  */
 
 export const Router = (props) => {
-  const ref = useRef(null);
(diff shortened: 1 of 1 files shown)


molefrog:
@cbbfcd Nice! I just realised that the `null` argument can be omitted here https://github.com/molefrog/wouter/blob/1adb0f5a5d3bcdf7b8e6186cc9b507a219bcc6ab/index.js#L63

The default ref value is `undefined` and it is going to pass the false check as good as `null`.

cbbfcd:
@molefrog Now the code size is a little less.

molefrog:
Nice work!

Issue #113, opened 2020-04-10
make static-location more useful for testing


davidje13:
Importing `wouter/static-location` is useful for server-side rendering, and also has some use when testing (e.g. ensuring the correct components are rendered for a given path). But it is lacking the ability to record changes to the URL (e.g. from redirects, links, or more generally `setLocation`).

Adding a small amount of state to this would make it more useful for testing, while maintaining its applicability to static rendering. In my own project, I have added the following, which could easily be included here as a replacement to `static-router`.

```ts
import type { LocationHook, Path } from 'wouter';

export interface StaticLocationHook extends LocationHook {
  locationHistory: Path[];
}

export default function staticLocationHook(
  path = '/',
  dynamic = false,
): StaticLocationHook {
  let latestLocation = path;

  const hook: StaticLocationHook = () => [
    dynamic ? latestLocation : path,
    (to: Path, replace?: boolean): void => {
      if (replace) {
        hook.locationHistory.pop();
      }
      hook.locationHistory.push(to);
      latestLocation = to;
    },
  ];

  hook.locationHistory = [path];
  return hook;
}
```

This allows testing like so:

```ts
const locationHook = staticLocationHook('/');

const dom = render((
  <Router hook={locationHook}>
    <MyComponent />
  </Router>
), { queries });

// (click a link)

expect(locationHook.locationHistory).toEqual(['/', '/my-new-path']);
```

And to perform transition tests, `dynamic` can be set true, causing the page to navigate realistically when links are clicked, etc.

molefrog:
@davidje13 Awesome work! This definitely looks like something we could add to the library.

cbbfcd:
add a fake 'history' 🚀

davidje13:
Want me to put together a PR? I'm guessing the necessary bits would be implementation (as shown here), typescript declaration, tests, and documentation updates in the readme?

molefrog:
@davidje13 Sure, I would appreciate you sending a PR. Let's discuss an API in details. Here are my thoughts.

- Since `staticLocation`'s main use case is SSR, it would be a bit risky to make the history recording functionality on by default. The recording is mostly needed in tests, so I assume something like this `staticLocationHook({ record = false })` should work.
[...]

PR #123, opened 2020-07-08
Add static location hook history [#113]

davidje13:
As discussed in #113. This adds `.history` to `staticLocationHook`.

This can be used for testing:

```js
const locationHook = staticLocationHook('/', { record: true });

const dom = render((
  <Router hook={locationHook}>
    <MyComponent />
  </Router>
), { queries });

// (click a link)

expect(locationHook.history).toEqual(['/', '/my-new-path']);
```

And can be used to perform automatic redirection when server-side rendering:

```js
const locationHook = staticLocationHook(requestedPage, { record: true });

const dom = render(<Router hook={locationHook}><App /></Router>);

const redirect = locationHook.history.slice(-1)[0];
if (redirect !== requestedPage) {
  return Redirect(redirect);
} else {
  return Ok(dom);
}
```

Notes:

- I didn't implement `dynamic`: when I looked at it properly, I found that it would need a lot more logic to make it work correctly with hooks (local state with detection of changes on the global state, similar to the existing code in `use-location.js`)
- This adds TypeScript definition tests to `static-location` which didn't exist before
- I noticed that this is not a drop-in replacement for the browser location hook because it doesn't implement `base`. I decided not to fix this since it is an existing unrelated bug (and fixing it would change behaviour whereas this PR maintains old behaviour unless the new option is specified).

More generally, I think `dynamic` and `base` should both be implemented by allowing more code to be shared from `use-location`. It feels like user-provided hooks could benefit from not having to worry about syncing global/local state. But I'm not sure what the best approach is there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,33 @@ const handleRequest = (req, res) => {
 
 Make sure you replace the static hook with the real one when you hydrate your app on a client.
 
+If you want to be able to detect redirects you can provide the `record` option:
+
+```js
+import { renderToString } from "react-dom/server";
+import { Router } from "wouter";
+import staticLocationHook from "wouter/static-location";
+
+import App from "./app";
+
+const handleRequest = (req, res) => {
+  const location = staticLocationHook(req.path, { record: true });
+  const prerendered = renderToString(
+    <Router hook={location}>
+      <App />
+    </Router>
+  );
+
+  // location.history is an array matching the history a
+  // user's browser would capture after loading the page
+
+  const finalPage = locationHook.history.slice(-1)[0];
+  if (finalPage !== req.path) {
+    // perform redirect
+  }
+};
+```
+
 ### 1KB is too much, I can't afford it!
 
 We've got some great news for you! If you're a minimalist bundle-size nomad and you need a damn simple
--- a/static-location.js
+++ b/static-location.js
@@ -1,4 +1,17 @@
 // Generates static `useLocation` hook. The hook always
 // responds with initial path provided.
 // You can use this for server-side rendering.
-module.exports = (path = "/") => () => [path, x => x];
+export default (path = "/", { record = false } = {}) => {
+  let hook;
+  const navigate = (to, { replace } = {}) => {
+    if (record) {
+      if (replace) {
+        hook.history.pop();
+      }
+      hook.history.push(to);
+    }
+  };
+  hook = () => [path, navigate];
+  hook.history = [path];
+  return hook;
+};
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -424,6 +424,33 @@ const handleRequest = (req, res) => {
...
+  const prerendered = renderToString(
+    <Router hook={location}>
+      <App />
+    </Router>
+  );
+
+  // location.history is an array matching the history a
+  // user's browser would capture after loading the page
+
+  const finalPage = locationHook.history.slice(-1)[0];
```
molefrog, line 447: I wonder if there more clear way to get the last element of the array. For some beginner readers of the README this might look confusing. 
I was thinking we could do that with array destructuring first, but then it appeared it's not possible with ES6. 

Any ideas?

molefrog:
> I didn't implement dynamic: when I looked at it properly, I found that it would need a lot more logic to make it work correctly with hooks (local state with detection of changes on the global state, similar to the existing code in use-location.js)

Oh well, I knew this was going to come up. One of hidden things in wouter I did because of the size limit was passing the entire `router` object to the hook. Router is just a plain object that never changes during the app lifecycle (unless you change the top-level options). That object holds all the configuration and ... it can be used to hold extra values, essentially acting as a mediator.

Here is what it allows to do:
```js
import { createNanoEvents } from 'nanoevents'

const memoryListener = Symbol()
const currentPath = Symbol()

const inMemoryHook = router => {
  router[memoryListener] = router[memoryListener] || createNanoEvents()
  
  // you now can use the listener now to notify other components about the state changes
}
```

README.md
davidje13: Well it could always be added as a dedicated property on the hook for convenience. Size doesn't really matter so much for this code because it's only going to be server-side / dev-side, but I didn't want to bloat the API. So if you'd prefer, it could be added as `locationHook.current` or something.

molefrog: I think it's going to do the job 😄 Let's merge the PR.

molefrog:
@davidje13 The PR is merged, I plan to make a new release soon and will mention you in the release notes. Nice work and thanks for contributing!

May I ask what kind of projects are you using wouter in?

davidje13:
Cool 🙂

> May I ask what kind of projects are you using wouter in?

Just dabbling with it for now. I swapped out react-router-dom for wouter in my hobby project [Refacto](https://github.com/davidje13/Refacto) (see [here](https://github.com/davidje13/Refacto/blob/master/src/frontend/src/components/App.tsx) and [here](https://github.com/davidje13/Refacto/blob/master/src/frontend/src/components/RetroRouter.tsx#L154)) when I was considering switching from react to preact (still haven't made the switch to preact though).

PR #129, opened 2020-08-02
Reuse & expose events

o-alexandrov:
We stumbled upon a need to react on the same events wouter's `useLocation`, but outside of a hook & react itself.
For example, to clean up certain state from all components on history's state changes, but not notifying them about it.
- basically a case similar to useRef, but outside of react

This PR should decrease the bundle a tiny bit (or probably leave it the same, due to additional export) for an already tiny library :D
And let users to reuse some of the events, keeping wouter as a source of truth.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,10 @@ import {
   createElement as h,
 } from "./react-deps.js";
 
+/**
+ * consts
+ */
+export { eventsHistory } from "./use-location";
 /*
  * Part 1, Hooks API: useRouter, useRoute and useLocation
  */
--- a/index.js
+++ b/index.js
@@ -12,10 +12,6 @@ import {
   createElement as h,
 } from "./react-deps.js";
 
-/**
- * consts
- */
-export { eventsHistory } from "./use-location";
 /*
  * Part 1, Hooks API: useRouter, useRoute and useLocation
  */
--- a/use-location.js
+++ b/use-location.js
@@ -1,5 +1,10 @@
 import { useEffect, useRef, useState, useCallback } from "./react-deps.js";
 
+const eventPopState = "popstate"
+const eventPushState = "pushstate"
+const eventReplaceState = "replaceState"
+export const eventsHistory = [eventPopState, eventPushState, eventReplaceState]
+
 export default ({ base = "" } = {}) => {
   const [path, update] = useState(currentPathname(base));
   const prevPath = useRef(path);
@@ -16,8 +21,7 @@ export default ({ base = "" } = {}) => {
       prevPath.current !== pathname && update((prevPath.current = pathname));
     };
 
-    const events = ["popstate", "pushState", "replaceState"];
-    events.map((e) => addEventListener(e, checkForUpdates));
+    eventsHistory.map((e) => addEventListener(e, checkForUpdates));
 
     // it's possible that an update has occurred between render and the effect handler,
     // so we run additional check on mount to catch these updates. Based on:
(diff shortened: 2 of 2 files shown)


use-location.js
```diff
@@ -1,5 +1,10 @@
 import { useEffect, useRef, useState, useCallback } from "./react-deps.js";
 
+const eventPopState = "popstate"
+const eventPushState = "pushstate"
```
molefrog, line 4:
```suggestion
const eventPushState = "pushState"
```

The `eventPushState` as a name of a method when performing a navigation with `history.pushState()` and `history.replaceState()`

use-location.js
```diff
@@ -1,5 +1,10 @@
...
+const eventReplaceState = "replaceState"
```
molefrog, line 5: Should we maybe rename these vars to uppercase? Like:
```suggestion
const REPLACE_STATE_EVENT = "replaceState"
```
?

use-location.js
```diff
@@ -1,5 +1,10 @@
...
+export const eventsHistory = [eventPopState, eventPushState, eventReplaceState]
```
molefrog, line 6: Was thinking about a shorter name for this. Maybe just `events`? Exported constant names can't be minified, so the length matters 😃

index.js
```diff
@@ -12,6 +12,10 @@ import {
   createElement as h,
 } from "./react-deps.js";
 
+/**
+ * consts
+ */
+export { eventsHistory } from "./use-location";
```
molefrog, line 18: While it seems convenient to export these, I really would not want to export implementation-specific location constants from `wouter`. The reason is that wouter can be customised with a variety of locations (hash-based, state-based, memory, static location for SSR and so on). Let's just make that `use-location.js` specific (which by the way should be named `use-browser-location.js` since it's really confusing with `useLocation` hook from `wouter`).

```js
import { events } from "wouter/use-location";

events.forEach(event => window.addEventListener(event, cb));
```

molefrog:
>  For example, to clean up certain state from all components on history's state changes, but not notifying them about it

Can you provide a more specific example? Just curious what the use case might be, maybe it can be solved by just using an API provided by wouter.

use-location.js
o-alexandrov: Could you please clarify why you think mangling won't work during minification?
In my experience and just checked some bundles minified with [terser](https://github.com/terser/terser) and don't see an issue w/ mangling.

use-location.js
o-alexandrov: While I understand you concern, I believe some users might have [eslint rule](https://eslint.org/docs/rules/camelcase) for the naming convention:
- "always"

We would make it a pain for them, while if the vars are camelCase, there is a less probability of pissing someone off :)

use-location.js
o-alexandrov: My bad, fixed it and added a comment for others to quickly check MDN, if additions/changes would be made

index.js
o-alexandrov: thank you, fixed it as you requested

o-alexandrov:
I believe requested changes are made. Please re-review when convenient.
About size-limit, unfortunately I am not familiar with using that library.

After brief familiarization with it, I see they support `custom` bundling configurations.
IMHO, if you would like to get the true size of the library in production of `use-location.js` file, then you should consider to set a custom `rollup + terser` configuration, so that way we could see the weight of it after minifaction.

As of now, the size increased a bit due to:
- naming affecting the size
- exporting a new variable

+2 bytes for the whole bundle
+18 bytes for use-location

but in reality, the size shouldn't increase due to mangling & other processing by terser

use-location.js
molefrog: Didn't think about that, makes sense!

molefrog:
Release in v2.5.2 https://github.com/molefrog/wouter/releases/tag/v2.5.2

PR #134, opened 2020-10-16
Add strict routes section to FAQ

### LINKED ISSUE #106: `strict` option on <Route>
See https://codesandbox.io/s/serene-lake-j5ul4. When routing to `/foo/`, the `/foo` route unexpectedly matches even thought a `/foo/` route exists. When one flips around the two `Route`s order, it works as expected:

```` jsx
<Route path="/foo/">At /foo/</Route>
<Route path="/foo">At /foo</Route>
````

This could be solved by either adding a [`strict`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Route.md#strict-bool) option like `react-router`, or just be strict by default, but that may be a breaking change for some users.

omgovich:
Closes #106

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,6 +450,30 @@ return (
 
 **[▶ Demo Sandbox](https://codesandbox.io/s/5zjpj19yz4)**
 
+### Are strict routes supported?
+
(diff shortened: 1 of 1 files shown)


molefrog (approved):
Awesome, thanks a lot!

Issue #136, opened 2020-10-19
`useLayoutEffect` handler in redirect should not return anything


thurinus:
I'm evaluating Wouter for a small Typescript-based React project, and am running into an issue when attempting a `Redirect` while hash-based routing is set up.

I created a [sandbox that replicates the problem](https://codesandbox.io/s/upbeat-knuth-c3pv4), lightly-edited from the example in the readme. This is the error that appears when the redirect link is clicked:

```
TypeError
func.apply is not a function
HTMLUnknownElement.callCallback
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:95:18
Object.invokeGuardedCallbackDev
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:119:20
invokeGuardedCallback
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:150:35
safelyCallDestroy
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:13242:9
eval
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:13590:25
unstable_runWithPriority
https://c3pv4.csb.app/node_modules/scheduler/cjs/scheduler.development.js:653:12
runWithPriority$1
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:7219:14
commitUnmount
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:13585:19
unmountHostComponents
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:13848:11
commitDeletion
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:13873:9
commitMutationEffects
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:15294:15
HTMLUnknownElement.callCallback
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:95:18
Object.invokeGuardedCallbackDev
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:119:20
invokeGuardedCallback
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:150:35
commitRootImpl
https://c3pv4.csb.app/node_modules/react-dom/cjs/react-dom.development.js:15112:13
unstable_runWithPriority
[...]

PR #137, opened 2020-10-20
fix: useLayoutEffect should return nothing

cbbfcd:
for [#136](https://github.com/molefrog/wouter/issues/136)

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,7 +153,7 @@ export const Redirect = (props) => {
 
   // empty array means running the effect once, navRef is a ref so it never changes
   // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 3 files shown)


omgovich:
Nice catch!

molefrog:
Great! @cbbfcd If possible, can you also write a simple spec that tests that the problem described in the issue has been fixed?
Steps to reproduce:
1. Write a custom hook that when `setLocation('/any')` is called returns a non-function (for example, "foo").
2. Provide the hook to `Router`, place a `Redirect` component.

With your fix, the spec should go green, and fail without. Thanks!

cbbfcd:
@molefrog 👌

molefrog (approved):
Looks great!

PR #138, opened 2020-10-20
feat: basepath should be case-insensitive

cbbfcd:
hello, @molefrog 

The [wouter `matcher` is case-insensitive by default](https://github.com/molefrog/wouter/blob/master/matcher.js#L65), so the `basepath` should be case-insensitive default too ?

keep `pathname` and` basepath` with the same default behavior, so we can reducing user confusion. 

Of course, we can also provide a `sensitive` parameter, but it should be applied to `matcher` and `basepath` at the same time, otherwise, we will ignore case by default.

Do you have any suggestions？

### DIFF
--- a/use-location.js
+++ b/use-location.js
@@ -9,7 +9,7 @@ const eventReplaceState = "replaceState";
 export const events = [eventPopstate, eventPushState, eventReplaceState];
 
 export default ({ base = "" } = {}) => {
-  const [path, update] = useState(currentPathname(base));
+  const [path, update] = useState(currPathname(base));
   const prevPath = useRef(path);
 
   useEffect(() => {
@@ -20,7 +20,7 @@ export default ({ base = "" } = {}) => {
     // unfortunately, we can't rely on `path` value here, since it can be stale,
     // that's why we store the last pathname in a ref.
     const checkForUpdates = () => {
-      const pathname = currentPathname(base);
+      const pathname = currPathname(base);
       prevPath.current !== pathname && update((prevPath.current = pathname));
     };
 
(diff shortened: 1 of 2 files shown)


use-location.js
```diff
@@ -20,7 +20,7 @@ export default ({ base = "" } = {}) => {
     // unfortunately, we can't rely on `path` value here, since it can be stale,
     // that's why we store the last pathname in a ref.
     const checkForUpdates = () => {
-      const pathname = currentPathname(base);
```
molefrog, line 23: Can we get the old name back, please 🙏 
`currentPathname` is just more readable

use-location.js
```diff
@@ -75,5 +75,5 @@ const patchHistoryEvents = () => {
   return (patched = 1);
 };
 
-const currentPathname = (base, path = location.pathname) =>
-  !path.indexOf(base) ? path.slice(base.length) || "/" : path;
+const currPathname = (base, path = location.pathname.toLowerCase()) =>
```
molefrog, line 78: Actually, it feels like the path doesn't have to be lowercased. The matcher will take care of it (you can test it as weel).
```suggestion
const currPathname = (base, path)) =>
```

molefrog (changes requested):

use-location.js
cbbfcd: Of course, I just want to reduce the number of bytes

use-location.js
cbbfcd: oh~,nice, can reduce some bytes, i will test it

cbbfcd:
@molefrog done. please review again.

use-location.js
cbbfcd: `location.pathname` should be lowercased, so we can slice the `location.pathname` with `base` successfully.
e.g.

```js
location.pathname = '/App/foo'
// and
base = '/app'
```

molefrog: Oh, you're right, my bad 😄

molefrog (approved):
Great work @cbbfcd!
The CI is failing, but that's because of a problem with codecov command. I'll fix that later.

Issue #142, opened 2020-10-22
Bug. "navigate" method of useLocation() sets title to 0


o-alexandrov:
`document.title` is set to `0` on some OS, for example: `iOS 13 and 14`.

The problem is in how we use `navigate`.
Both [pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) and [replaceState](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) do set the title as documentation indicates in some browsers.

The documentation suggests to use an `""` (empty string) instead of currently used `0` (Number).

PR #143, opened 2020-10-22
[WIP] useNavigate (to not listen to history's state)

### LINKED ISSUE #140: Expose navigate, not requiring to subscribe to location state
Currently to manage location, `wouter` doesn't make it trivial for the `user (developer)` to understand:
- there is no need to subscribe to `location` state via `useLocation()`

There are many cases, when updating `history`'s state via `history`'s (`pushState` and `replaceState`) is more efficient than subscribing to `location` state via `useLocation()` and then updating state via second array's item (typically named as `navigate`) of the returned value.

I propose we:
1. get rid of optimization flow, where we create a [memoized callback (line 42)](https://github.com/molefrog/wouter/blob/master/use-location.js#L42)
    - instead we declare `navigate` outside of the `useLocation` scope
2. expose the `navigate` callback to be exported without necessity to subscribe to `location` state
3. make `useLocation()` to return `const location = useLocation()`
    - there would be no need to return an array `const [location, navigate] = useLocation()`

Please let me know, if you are open to this, so I'd know whether it makes sense to create a PR for it. Thank you.
It should:
- decrease the bundle size a tiny bit
- allow to easily create components that change history, but don't re-render on history updates

o-alexandrov:
Fixes #140 #142

I haven't updated typings yet.
I will do so later. Please let me know if there is something you would like to change in the PR.

It's a breaking change, but brings a very important feature to have a stateless `navigate` method.

**Please be informed, this PR contains changes from** #141 . Since that PR is approved, to make my life easier, I include changes from that branch also.

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,8 @@ const buildRouter = ({
   hook = locationHook,
   base = "",
   matcher = makeMatcher(),
-} = {}) => ({ hook, base, matcher });
+  useNavigate
+} = {}) => ({ hook, base, matcher, useNavigate });
 
 export const useRouter = () => {
   const globalRef = useContext(RouterCtx);
@@ -36,6 +37,9 @@ export const useRouter = () => {
   return globalRef.v || (globalRef.v = buildRouter());
 };
 
+// if you would like to customize, for example:
+//   - to manage state or document.title
+// override useNavigate through "Router" & don't use this default "useNavigate" Hook
 export const useNavigate = () => {
   const { base } = useRouter();
   return useCallback((to, { replace } = {}) =>
@@ -54,8 +58,9 @@ export const useRoute = (pattern) => {
 
 // internal hook used by Link and Redirect in order to perform navigation
 const useNavigateInternal = (options) => {
+  const { useNavigate: useOverride } = useRouter();
   const navRef = useRef();
-  const navigate = useNavigate();
+  const navigate = (useOverride || useNavigate)(); // it shouldn't break the rules of Hooks: "Don't call conditionally" because when override is applied, the condition is always true or false
 
   navRef.current = () => navigate(options.to || options.href, options);
   return navRef;
--- a/index.js
+++ b/index.js
@@ -1,4 +1,4 @@
-import locationHook from "./use-location.js";
+import locationHook, { eventPushState, eventReplaceState } from "./use-location.js";
 import makeMatcher from "./matcher.js";
 
 import {
@@ -36,20 +36,26 @@ export const useRouter = () => {
   return globalRef.v || (globalRef.v = buildRouter());
 };
 
+export const useNavigate = () => {
+  const { base } = useRouter();
+  return useCallback((to, { replace } = {}) =>
+    history[replace ? eventReplaceState : eventPushState](null, "", base + to), [base]);
+};
+
 export const useLocation = () => {
   const router = useRouter();
   return router.hook(router);
 };
 
 export const useRoute = (pattern) => {
-  const [path] = useLocation();
+  const path = useLocation();
   return useRouter().matcher(pattern, path);
 };
 
 // internal hook used by Link and Redirect in order to perform navigation
-const useNavigate = (options) => {
+const useNavigateInternal = (options) => {
   const navRef = useRef();
-  const [, navigate] = useLocation();
+  const navigate = useNavigate();
 
   navRef.current = () => navigate(options.to || options.href, options);
   return navRef;
@@ -89,7 +95,8 @@ export const Route = ({ path, match, component, children }) => {
 };
 
 export const Link = (props) => {
-  const navRef = useNavigate(props);
+  useLocation(); // subscribe to history's state
+  const navRef = useNavigateInternal(props);
   const { base } = useRouter();
 
   let { to, href = to, children, onClick } = props;
@@ -125,7 +132,7 @@ export const Link = (props) => {
 
 export const Switch = ({ children, location }) => {
   const { matcher } = useRouter();
-  const [originalLocation] = useLocation();
+  const originalLocation = useLocation();
 
   children = Array.isArray(children) ? children : [children];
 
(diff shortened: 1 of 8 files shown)


cbbfcd:
Great change, but it seems to limit the free expansion space, through `hook`

o-alexandrov:
@cbbfcd please clarify what do you mean by: `free expansion space`

o-alexandrov:
@molefrog could you kindly review & merge other 2 PRs #141 #144 ?
- it would help me to clean up this PR

cbbfcd:
I understand your core idea is to strip `useNavigate` out, it's great idea.

The previous method of `useLocation` has an advantage, allowing users to expand through `hook` props in `Router`, expand the way of `navigate` or expand the way of monitoring `path`, and so on. this can improve the ecology of `wouter` through the addition of `hook`, just as support `search`, support `hash`, support `pass a state` when navigate.

so, 

* may be [useNavigate](https://github.com/molefrog/wouter/pull/143/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R39) shoule be added in `Router` props?

```js
const buildRouter = ({
  hook = locationHook,
  base = "",
  matcher = makeMatcher(),
  // here
  useNavigate,
} = {}) => ({ hook, base, matcher });
```

o-alexandrov:
Thank you for the clarification.
You are right, I’ll modify the buildRouter, so you could customize the navigate method to add:
- state management
- or any other custom replacement

We shouldn’t pass the useNavigate, just the inner navigate is enough.
Let me please make a change later today and I’ll ping you here

omgovich:
Hi @o-alexandrov. As far as I understand the initial idea was to make the library easier to use.
Maybe it's just me, but I don't see how these changes improve DX or make the library significantly easier.

We need to have enough reasons to make a breaking API change.

o-alexandrov:
@omgovich hi, I don't think we should discuss DX when the result of this library is screens with freezing on mobile devices, when you do animation while switching between routes.
- it happens, because components subscribe to history's state, resulting in reevaluation & sometimes unnecessary rerenders

I disagree separation of `navigate` and `location` makes the library more complex.
It is the opposite, it makes DX easier, reasons:
- many react devs are used to `react-router` (the most famous routing lib) or `@reach/router` (2nd most famous)
  - both of them have a separate `useLocation` and `useHistory`/`useNavigate`

**So why do you think the decision in `wouter` to introduce a new pattern was simpler, than conforming to a common one?**

Of course there are much more reasons to support the separation, but I hope the above makes it an obvious decision.

o-alexandrov:
@cbbfcd I've thought about how we could handle a case:
- when user (developer) wants to override `useNavigate`

It was actually as simple as you proposed 👍 , please review this [last commit](https://github.com/molefrog/wouter/pull/143/commits/e180bfa88931994614d5af2f40ede828dc658e7f)

cbbfcd (approved):
great job!

o-alexandrov:
@cbbfcd Thank you, this PR still needs:
- approval of the general proposal from @molefrog , whether Alexey agrees to split `navigation` & `location`
- I need to fix typings (tests are fixed and all passing, but needs coverage check, I will do it)

I postpone remaining cleanups until we know the overall change would be accepted by @molefrog

molefrog:
Thanks guys, I need time to review the code. I wrote my ideas in the original issue.

index.js
```diff
@@ -36,20 +37,30 @@ export const useRouter = () => {
   return globalRef.v || (globalRef.v = buildRouter());
 };
 
+// if you would like to customize, for example:
+//   - to manage state or document.title
+// override useNavigate through "Router" & don't use this default "useNavigate" Hook
+export const useNavigate = () => {
+  const { base } = useRouter();
+  return useCallback((to, { replace } = {}) =>
+    history[replace ? eventReplaceState : eventPushState](null, "", base + to), [base]);
```
molefrog, line 46: This will mess up the separation of concerns: wouter itself should know nothing about the implementation details of the location interface. It'll break existing apps as well: those who use custom locations hooks, for example hash-based or internal (global state) routing.

static-location.js
```diff
@@ -1,17 +1,4 @@
...
-  const navigate = (to, { replace } = {}) => {
-    if (record) {
-      if (replace) {
-        hook.history.pop();
-      }
-      hook.history.push(to);
-    }
-  };
-  hook = () => [path, navigate];
-  hook.history = [path];
```
molefrog, line 15: We need to figure out how to add support for recording changes back. I know that the signature of `useLocation` is different now, but there should be a way for people to have proper testing support. See https://github.com/molefrog/wouter/issues/113

molefrog (commented):
Thanks for sending the PR! I've taken my time to study the code and it feels like the idea is great (providing a subscription-independent  `useNavigate` hook), but the implementation could be improved. I would suggest for us to take a few steps back and carefully design the new API first. 

Here is what I think would work:
- The `useLocation()` signature should stay the same for now (return an array). I would hold it off for the next major release.
- The new hook `const navigate = useNavigate()` is exported from "wouter"
- The hook takes the `useNavigate` implementation from the location module. 
- The standard location `use-location.js` based on History API, somehow exports `useNavigate`
- If a custom hook **does not** export `useNavigate` -> then it should fall back to using `useLocation()`.
- Probably a new abstraction called History is needed. A history is a simple object that tells how to subscribe to location changes and how to change the locaiton.

I'll write more detailed API examples in the original issue.

o-alexandrov:
I apologize, but I am not interested in supporting old `useLocation`'s pattern returning an array `[location, navigate]`.
I believe `wouter` should make a breaking change in API and release it as a new major version.

You are also welcome to take everything from this PR, adjust it to your needs, and/or close the PR how you see fit.
For anyone interested in using this PR, you are welcome to install from the github repo:
```json
{
  "dependencies": {
    "wouter": "o-alexandrov/wouter#feat/navigate"
  }
}
```

PR #144, opened 2020-10-23
fix: passing state & title to pushState & replaceState

### LINKED ISSUE #142: Bug. "navigate" method of useLocation() sets title to 0

o-alexandrov:
Fixes #142 

As #142 describes, this PR corrects `state` & `title` params for `pushState` & `replaceState` to the recommended `falsy` ones `null` for `state` & `""` for `title`

### DIFF
--- a/test/preact.test.js
+++ b/test/preact.test.js
@@ -8,7 +8,7 @@ const { Route, Link, Switch } = require("../index.js");
 
 describe("Preact support", () => {
   beforeEach(() => {
-    history.replaceState(0, 0, "/");
+    history.replaceState(null, "", "/");
     setupRerender();
   });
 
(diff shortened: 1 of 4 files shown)


molefrog (approved):
Thanks!

PR #156, opened 2020-12-21
feat: take Fragments and Arrays into account for Switch

### LINKED ISSUE #154: Switch component breaks with fragments
## About

Hey there! Thanks for the awesome library!

There seems to be an issue with the switch component when fragments are used.

## Reproduction

 - Create a switch component with some child routes.
 - Wrap some of the child routes in a fragment (code below)
 - Add more routes outside of the fragment
 - Load the page on a route that doesn't match any of the routes in the fragment

```tsx
<Switch>
  <>
    {literalRoutes.map((route) => (
        <Route key={route.url} path={route.url}>
          <AppRoute {...route} />
        </Route>
     ))}
  </>   
  <Route key={'fallback'}>
    <Redirect to={'/'} />
   </Route>,
</Switch>
```

### Expected result

Wouter renders the first matching route (or fallback route)

### Actual result

Wouter gives up parsing `Switch` child components after parsing the fragment and renders nothing.

## Additional info

The following example is what I'm using as a workaround for now

```tsx
<Switch>
  {[
    ...literalRoutes.map((route) => (
      <Route key={route.url} path={route.url}>
        <AppRoute {...route} />
      </Route>
    )),
    <Route key={'fallback'}>
      <Redirect to={'/'} />
    </Route>,
  ]}
</Switch>
```

jvdsande:
This PR fixes #154.

I open it in case the ideas discussed in #154 are accepted. This adds a `flattenChildren` method used inside the `Switch` component to gracefully handle Arrays and Fragments as children of Switch. This opens up nicer patterns for static fallback routes with dynamically defined siblings.

This PR also adds two tests for the Switch component, validating its handling of Arrays and Fragments.

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,9 +143,7 @@ export const Switch = ({ children, location }) => {
   const { matcher } = useRouter();
   const [originalLocation] = useLocation();
 
-  children = flattenChildren(children);
-
-  for (const element of children) {
+  for (const element of flattenChildren(children)) {
     let match = 0;
 
     if (
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ import {
   isValidElement,
   cloneElement,
   createElement as h,
+  Fragment,
 } from "./react-deps.js";
 
 /*
@@ -128,11 +129,21 @@ export const Link = (props) => {
   return cloneElement(jsx, extraProps);
 };
 
+const flattenChildren = (children) => {
+  return Array.isArray(children)
+    ? children.flatMap((c) =>
+        c.type === Fragment
+          ? flattenChildren(c.props.children)
+          : flattenChildren(c)
+      )
+    : [children];
+};
+
 export const Switch = ({ children, location }) => {
   const { matcher } = useRouter();
   const [originalLocation] = useLocation();
 
-  children = Array.isArray(children) ? children : [children];
+  children = flattenChildren(children);
 
   for (const element of children) {
     let match = 0;
--- a/index.js
+++ b/index.js
@@ -131,10 +131,12 @@ export const Link = (props) => {
 
 const flattenChildren = (children) => {
   return Array.isArray(children)
-    ? children.flatMap((c) =>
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -128,11 +129,21 @@ export const Link = (props) => {
...
+};
+
 export const Switch = ({ children, location }) => {
   const { matcher } = useRouter();
   const [originalLocation] = useLocation();
 
-  children = Array.isArray(children) ? children : [children];
+  children = flattenChildren(children);
 
   for (const element of children) {
```
molefrog, line 148:
```suggestion
  for (const element of flattenChildren(children)) {
```

jvdsande: Ah yes indeed :) Shaves 2 bytes off the final build. Applied!

index.js
```diff
@@ -128,13 +129,21 @@ export const Link = (props) => {
   return cloneElement(jsx, extraProps);
 };
 
+const flattenChildren = (children) => {
+  return Array.isArray(children)
+    ? children.flatMap((c) =>
```
molefrog, line 134: I wonder if it is safe to use `flatMap` here. I know it's widely supported, but it's part of ES2019. wouter currently targets ES2015, would that maybe better to use a short polyfill: `var flatMap = (arr, cb) => [].concat(...arr.map(cb))`

molefrog: But overall, this is quite smart! 👍

jvdsande: Ah yes, did not think of that, so used to Typescript/ESNext, I always struggle in build-less environments.
I'll add the polyfill you suggested tomorrow. It will increase the size by a few bytes but not that much I guess.

jvdsande: I've added the polyfill. I inlined it in `flattenChildren` directly as this was best for bundle size. Bundle size increase was 5 bytes for the polyfill.

molefrog (approved):

molefrog:
Great work!

PR #159, opened 2020-12-26
Ignore external routing

### LINKED ISSUE #158: Disable interception of external URLs
## About

When an `onClick` is called on a link which is external, it would be useful if wouter didn't intercept the event.

This prevents the following error at runtime

> Uncaught DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'https://some.external/link' cannot be created in a document with origin 'https://current.origin' and URL 'https://current.origin/url'.

### Example

Here's an example where this would be useful

```tsx
import { Link } from 'wouter';

const navigation = [
  { text: "Internal link", url: "/some-url" } ,
  { text: "External link", url: "https://mysite.com" } ,
]

export const Navigation = () => (
  <nav>
    {navigation.map(({ text, url }) => <Link href={url}>{text}</Link>)}
  </nav>
)
```

andyrichardson:
Fix #158

Edit: Updated to remove [URL](https://caniuse.com/url) 

Todo: 
 - Add test
 - Get feedback from @molefrog as to whether this will get merged

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,6 +107,13 @@ export const Link = (props) => {
         event.button !== 0
       )
         return;
+      
+      // ignores the navigation when routing is outside
+      // of application
+      const target = new Url(event.target.href, window.location.origin);
+      if (target.host !== window.location.host) {
+        return;
+      }
 
       event.preventDefault();
       navRef.current();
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -107,6 +107,13 @@ export const Link = (props) => {
         event.button !== 0
       )
         return;
+      
+      // ignores the navigation when routing is outside
+      // of application
+      const target = new Url(event.target.href, window.location.origin);
```
andyrichardson, line 113: That second argument is ignored if the URL contains the host

Screen Shot 2020-12-26 at 13 01 03

PR #163, opened 2021-01-02
Simplify code example of hash-based routing

joaopaulobdac:
The `navigate` function can be moved outside the function component since it doesn't have any dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ If you do need a custom history observer, for example, for hash-based routing, y
 As an exercise, let's implement a simple location hook that listens to hash changes:
 
 ```js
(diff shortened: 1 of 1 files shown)


molefrog (approved):
Not as optimized as it was before, but I guess it really doesn't matter in the README :)

joaopaulobdac:
:thinking:  What do you mean? I don't understand how declaring the function inside the component would be better optimized

molefrog:
You're right, my bad! The function is declared globally, so the reference will be the same between re-renders. `useCallback` does absolutely the same, except that it allows to have functions bound to the local context, but in this case `navigate` is a standalone function, so declaring it globally is way better 👍

Issue #58, opened 2019-07-17
Support for search


worudso:
In the latest version of wouter, when the location is `somewebsite.com/users?name=john`
```typescript
const [location] = useLocation();
```
`location` only retrieves `/users` emitting `?name=john`. And even though I push new query params with other methods, re-rendering is not triggered. But sometimes a view needs to depend on search params.

I think we have two options in a broad sense
- add `useSearch` thing
  ```typescript
  const [search, setSearch] = useSearch();
  search // "?name=john"
  ```
  or we can let wouter give `searchParams` as a `URLSearchParams` object, the parsed one.
  or how about `useURL` which gives current location as URL object?
- let `location` be `location.pathname + location.search`.
  > breaking change!

  In this case, we got a new problem to redefine routing rules. Should `useRoute` provide matching for search parameters? I don't think so. Routing rules can go the same just ignoring search.
  And the parsing for `location.pathname + location.search` is annoying. AFAIK, there is no builtin function to parse it. And even if there is, the code to parse would be repeated in many components.

molefrog:
@worudso @dpr-dev Hey guys, thanks for the issue. 
Been thinking about this a lot and what I had in mind is following:
```js
// useLocation hook may return a third argument, which is basically a hash 
// of query params used; this is completely optional and apps that use 
// the hook already won't break
const [location, setLocation, queryParams] = useLocation();

// the signature of the useRoute hook stays the same, however it may return query params
// in the params hash along with dynamic segments
const [match, params] = useRoute("/app/users/:id");

// example: /app/users/alex?tab=settings
// -> { id: "alex", tag: "settings" }
```

In my opinion this can be one of the optimal solutions because:
1. Doesn't break the existing API
2. The query + route params merging technique is widely used, for example in Rails or Express (I guess).

worudso:
This is my fork
https://github.com/ownerclan/wouter/blob/master/use-search-params.js
And I implemented `useSearchParams` for my immediate needs.

I'm not saying we'd better go `useSearchParams`. Instead, I'm saying @molefrog 's new `useLocation` might be the right choice. I hope my experiences that dealt with search params help you make a better decision.

The current version of my `useSearchParams` returns an array of `[key, value]` tuples. But for convenience, returning an object with keys of params is best. I had used that version of `useSearchParams` at first. It's like react's old `state` and `setState` which allowed partial update of a state. This design had made my code looks better.

[...]

PR #164, opened 2021-01-02
Compatibility with react-use useSearchParam

joaopaulobdac:
[I commented about this](https://github.com/molefrog/wouter/issues/58#issuecomment-753407170) on #58 , so I thought I should open a PR to fix the problem I ran into. Currently, wouter doesn't have support for search params so I tried to use the [`useSearchParam` hook from react-use](https://github.com/streamich/react-use/blob/master/src/useSearchParam.ts), problem is that wouter and react-use added listeners for different events, making them incompatible since `useSearchParam` doesn't update when wouter does.

Plus, this change also follows the naming convention of window events.

### DIFF
--- a/use-location.js
+++ b/use-location.js
@@ -4,9 +4,9 @@ import { useEffect, useRef, useState, useCallback } from "./react-deps.js";
  * History API docs @see https://developer.mozilla.org/en-US/docs/Web/API/History
  */
 const eventPopstate = "popstate";
-const eventPushstate = "pushstate";
-const eventReplacestate = "replacestate";
-export const events = [eventPopstate, eventPushstate, eventReplacestate];
+const eventPushState = "pushState";
+const eventReplaceState = "replaceState";
+const events = [eventPopstate, eventPushState, eventReplaceState];
 
 export default ({ base = "" } = {}) => {
   const [path, update] = useState(() => currentPathname(base)); // @see https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
@@ -22,14 +22,14 @@ export default ({ base = "" } = {}) => {
       prevPath.current !== pathname && update((prevPath.current = pathname));
     };
 
-    events.map((e) => addEventListener(e, checkForUpdates));
+    events.map((e) => addEventListener(e.toLowerCase(), checkForUpdates));
 
     // it's possible that an update has occurred between render and the effect handler,
     // so we run additional check on mount to catch these updates. Based on:
     // https://gist.github.com/bvaughn/e25397f70e8c65b0ae0d7c90b731b189
     checkForUpdates();
 
-    return () => events.map((e) => removeEventListener(e, checkForUpdates));
+    return () => events.map((e) => removeEventListener(e.toLowerCase(), checkForUpdates));
   }, [base]);
 
   // the 2nd argument of the `useLocation` return value is a function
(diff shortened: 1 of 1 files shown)


joaopaulobdac (commented):
:man_facepalming: The event strings are not just used to add and remove event listeners, but also to call the events. So this means we would call `replacestate` instead of `replaceState` on the `history` API...

joaopaulobdac:
@molefrog I discarted the commit that changed the event names declarations and made it so that they are transformed to lowercase when registering listeners. Do you think this is ok? This way `replaceState` is registered as `replacestate`.

use-location.js
```diff
@@ -6,7 +6,7 @@ import { useEffect, useRef, useState, useCallback } from "./react-deps.js";
 const eventPopstate = "popstate";
 const eventPushState = "pushState";
 const eventReplaceState = "replaceState";
-export const events = [eventPopstate, eventPushState, eventReplaceState];
```
joaopaulobdac, line 9: I also remove this export statement because it isn't being used by any other file

joaopaulobdac:
The `useLocation` hook from react-use also does this:

https://github.com/streamich/react-use/blob/206f4b037109bcb8325a156edfe3ce6795b7976e/src/useLocation.ts#L4-L18

molefrog:
Hey @joaopaulobdac, I don't think the names of these events really matter. The events that wouter emits are virtual and non-standard: there is no built-in way to subscribe to location changes, so wouter matches native methods and triggers custom events. If both wouter and react-use patch the History API, they should play along well. 

Take a look at https://github.com/molefrog/wouter/pull/169, I have a feeling that the reason wouter didn't react to search changes was that it applied caching that ignore the search query.

use-location.js
molefrog: Well, this is intended, although not documented. Some apps use the name of the events in order to subscribe to their own location changes. I'm thinking about removing it in the next major release though.

joaopaulobdac:
> If both wouter and react-use patch the History API, they should play along well.

I think that one assignment overrides another. If I do `history[method] = 'a'` and then `history[method] = 'b'`, the first assignment would be gone, no?

use-location.js
joaopaulobdac: Ah ok. Thanks for clarifying.

molefrog:
> > If both wouter and react-use patch the History API, they should play along well.
> 
> I think that one assignment overrides another. If I do `history[method] = 'a'` and then `history[method] = 'b'`, the first assignment would be gone, no?

Not really, the original method (prior to patching) is always called with the same arguments before an event is triggered.
react-use does it here: https://github.com/streamich/react-use/blob/206f4b037109bcb8325a156edfe3ce6795b7976e/src/useLocation.ts#L9

and wouter does it here: https://github.com/molefrog/wouter/blob/master/use-location.js#L64

In the end, it creates a chain of hooks, so both wouter and react-use events should be fired. The worst thing that could happen is that they could have identical names for the events, but that's fine: the app will be rendered twice which isn't critical in case of a React app.

molefrog:
There are some weak points that patching has, well described in this issue: https://github.com/molefrog/wouter/issues/167

However, these are mostly performance-related concerns.

joaopaulobdac:
Great, I'll close this since it seems that with #169 we can react to changes to search params with `useLocation`

molefrog:
@joaopaulobdac There is a new version that I've published v2.7.2, hope it helps to resolve the problem.

PR #182, opened 2021-03-08
Add correct return type to `makeRegexpFn`

### LINKED ISSUE #180: Type issue with custom matcher
I'm running into an issue implementing a custom matcher. I've followed the example in the README but TS complains about the type.

The `makeRegexp` function passed to `makeCachedMatcher` (or `makeMatcher`) is expecting a return of type `RegExp`, but the example outlines `{ keys: Key[], regexp: RegExp }`.

```jsx
const convertPathToRegexp = (path: string) => {
  let keys: Key[] = [];

  const regexp = pathToRegexp(path, keys, { strict: true });
  return { keys, regexp };
};

const strictMatcher = makeCachedMatcher(convertPathToRegexp);
```

results in

```
Argument of type '(path: string) => { keys: Key[]; regexp: RegExp; }' is not assignable to parameter of type '(pattern: string, keys?: { name: string | number; }[] | undefined) => RegExp'.
  Type '{ keys: Key[]; regexp: RegExp; }' is missing the following properties from type 'RegExp': exec, test, source, global, and 13 more.ts(2345)
```

Am I missing something here, or is this an issue? I'm happy to raise a PR if it's as simple as changing the return type of `makeRegexp` 👍🏼 

Thanks for your work on this package too! It's really great and a nice light alternative to `react-router` with a flexible API.

shannonrothe:
This PR fixes the return type of `makeRegexpFn` which is passed to `makeMatcher`. Let me know if it requires changes 👍🏼 

Fixes #180

### DIFF
--- a/types/matcher.d.ts
+++ b/types/matcher.d.ts
@@ -20,14 +20,13 @@ export type Match<T extends DefaultParams = DefaultParams> =
 
 export type MatcherFn = (pattern: Path, path: Path) => Match;
 
-export type MakeRegexpFn = {
+export type PatternToRegexpResult = {
   keys: Array<{ name: string | number }>;
-  regexp: RegExp;
 };
 
 export default function makeMatcher(
   makeRegexpFn?: (
     pattern: string,
     keys?: Array<{ name: string | number }>
-  ) => MakeRegexpFn
+  ) => PatternToRegexpResult
 ): MatcherFn;
--- a/types/matcher.d.ts
+++ b/types/matcher.d.ts
@@ -22,11 +22,9 @@ export type MatcherFn = (pattern: Path, path: Path) => Match;
 
 export type PatternToRegexpResult = {
   keys: Array<{ name: string | number }>;
+  regexp: RegExp;
 };
 
 export default function makeMatcher(
-  makeRegexpFn?: (
-    pattern: string,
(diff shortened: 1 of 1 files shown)


types/matcher.d.ts
```diff
@@ -20,9 +20,14 @@ export type Match<T extends DefaultParams = DefaultParams> =
...
 
+export type MakeRegexpFn = {
+  keys: Array<{ name: string | number }>;
+  regexp: RegExp;
+};
+
 export default function makeMatcher(
   makeRegexpFn?: (
     pattern: string,
     keys?: Array<{ name: string | number }>
```
molefrog, line 31: The 2nd argument is obsolete, we can safely remove it. `keys` are returned instead.

types/matcher.d.ts
```diff
@@ -20,9 +20,14 @@ export type Match<T extends DefaultParams = DefaultParams> =
 
 export type MatcherFn = (pattern: Path, path: Path) => Match;
 
+export type MakeRegexpFn = {
```
molefrog, line 23: The type looks good. I would just rename it to 
```suggestion
export type PatternToRegexpResult = {
```

molefrog (changes requested):

molefrog (approved):

PR #183, opened 2021-03-10
Replace map with forEach

devrnt:
Since the result of `Array.map` in `use-location.js` isn't used it can be replaced, or is there a reason that I'm missing? Also in general `forEach` has better "performance" than `map`.

 Also I was going through the output bundle and I saw that the comments are still there, again am I missing something or aren't the comments stripped from the output bundles?

### DIFF
--- a/use-location.js
+++ b/use-location.js
@@ -31,14 +31,14 @@ export default ({ base = "" } = {}) => {
       }
     };
 
-    events.map((e) => addEventListener(e, checkForUpdates));
+    events.forEach((e) => addEventListener(e, checkForUpdates));
 
     // it's possible that an update has occurred between render and the effect handler,
     // so we run additional check on mount to catch these updates. Based on:
     // https://gist.github.com/bvaughn/e25397f70e8c65b0ae0d7c90b731b189
     checkForUpdates();
 
-    return () => events.map((e) => removeEventListener(e, checkForUpdates));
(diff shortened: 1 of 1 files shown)


molefrog:
> Since the result of Array.map in use-location.js isn't used it can be replaced, or is there a reason that I'm missing? 
No particular reason, since as you pointed out we don't care about the return value. It's just `map` is shorter than `forEach`. I don't mind changing that, as it could be better in terms of code readability. 

> Also in general forEach has better "performance" than map.
Well, it doesn't really matters, since in order to have the significant performance difference you need to iterate over a large array. But in our case it's always going to be 3 element, so the performance here doesn't matter. 

> Also I was going through the output bundle and I saw that the comments are still there, again am I missing something or aren't the comments stripped from the output bundles?
If by bundle you mean the code in the `cjs` folder then the short answer is: it's doesn't have to be minified, because it is responsibility of the bundler of the building party. You might want to keep comments in your dev build, but have them striped out in the production build. I also would like to point out that the `cjs` files are only used by Node and some outdated bundlers, since most of the modern build tools can understand the original ES6 module format (with import/exports).

molefrog (approved):

devrnt:
Indeed I forgot to mention "(to what extent it is applicable here)" in the "performance" part.

The `cjs` part of the answer  makes sense indeed, I thought that only an output of `cjs` format would be available to the end user (which wouldn't make sense).

PR #184, opened 2021-03-10
Chore/GitHub size workflow

devrnt:
### Add github size workflow

The size action will look at the `size-limit` entry in `package.json`, and report on every pull request, see [example](https://github.com/devrnt/react-use-wizard/pull/25#issuecomment-779427633) .

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -1,0 +1,12 @@
+name: Size
+on: [pull_request]
+jobs:
+  size:
+    runs-on: ubuntu-latest
+    env:
+      CI_JOB_NUMBER: 1
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


molefrog:
@devrnt It seems like the automation is failing. Can you investigate why?

devrnt:
Yes, it's because this action compares this fork the `master` branch. Since this action requires `size-limit` to be installed, and because it isn't in `master`, it's failing.

See action logs:
```
Branch 'master' set up to track remote branch 'master' from 'origin'.
/usr/local/bin/npm install
npm WARN @size-limit/file@4.10.1 requires a peer of size-limit@4.10.1 but none is installed. You must install peer dependencies yourself.
...
```

Anyway the `size-limit` boundaries should be increased a bit since they are failing now.

PR #211, opened 2021-09-06
TypeScript: Infer params from path

itsMapleLeaf:
Closes #210 

This will probably be a breaking change for TS users who manually typed their path params, I'm not sure if that warrants a major 🤔 Open to alternative approaches as well.

PR #217, opened 2021-09-26
Ts infer params from path use route

graphman65:
This is based on the work done on #211 
The new changes are:
- Make path params type inference backward compatible
- Make path params type inference work with `useRoute`
Existing code shouldn't break with these changes.


graphman65:
Has been included in #211 .
Closing.

PR #221, opened 2021-10-12
Add support for older version of typescript

### LINKED ISSUE #218: Smart-params type: Preact support
In https://github.com/molefrog/wouter/pull/211 a good work has been done to make the `Route` component as well as `useRoute` hook smart enough to automatically extract the proper params type from the path pattern. Now, in order to be able to release it we need to add a similar support to Preact version of the types.

graphman65:
Add supports for older versions of typescript.
#217 added param type inference from path. This feature is only supported by Typescript >= 4.1.
This PR is adding support for both the new versions and older versions of typescript.

It also add the missing types and tests for preact.

Also resolves #218


cbbfcd:
👍

molefrog:
Hi @graphman65, sorry I was having a 3 weeks off, just got back. The proposed change looks great, I was first confused with the approach of having two almost identical type sets for different TS versions and was hoping that there is some shortcut like [preprocessor conditions](https://github.com/Microsoft/TypeScript/issues/4691), but seems like the only proper way is indeed through `typesVersions`. I have some plans on rewriting the library in TS in the next major release, since supporting types separately is becoming a chore. I wonder how would we support two versions in that case.

molefrog (approved):
Looks good! However I would appreciate if someone else could review this as well. This is quite a change.

graphman65:
@molefrog 
Should we publish a new alpha version with the changes?

molefrog:
Yes, sorry, the `alpha.2` is live now.

molefrog:
So, has anyone tested this released on a real project yet?

graphman65:
Hey @molefrog,
I personally tested the react and preact versions on my projects.
I also tried using older versions of typescript and it switched to the correct types successfully.

Still would be good if someone else could test the new release as I'm probably not using all the features.

Issue #225, opened 2021-11-29
Base path inconsistencies


duarten:
I found a couple of situations where the handling around the base path isn't straightfoward:

1. When composing `<Router />`'s, like such:
    ```tsx
    <Router base="/a">
       <Router base="/b" />...</Router>
       ...
     </Router>
    ```
    The inner router's base will be `/b` instead of `/a/b`;
2. When using a `setLocation` obtained in another context, like:
    ```tsx
    const [, navigate] = useLocation()
    <Router base="/a">
         <Route>() => { navigate("/b") }</Route>
    </Router>
    ``` 
    This will navigate to `/b` instead of to `/a/b`.

Do you think these should be fixed? Fixing 1. is straightforward, fixing 2. can probably be done by keeping a stack of routers in the `RouterCtx`.

cbbfcd:
👍

HansBrende:
This single feature is pretty much the only feature react-router has on you all. Just beware of the stack of routers in the context... you don't want the app re-rendering every time they change. Instead, use a stable `ref` and call `ref.current` or `ref.current.parent.current` (to go backwards) etc. inside `useNavigate`, since really does not need to trigger a re-render or have a callback dependency on the router structure. See, e.g., the problems react router has run into over `useNavigate` (react router issue no. 7634).

molefrog:
Fixed via #265

PR #226, opened 2021-12-02
feat: propagate base paths across nested Routers

duarten:
Refs #225

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,17 +60,26 @@ const useNavigate = (options) => {
  * Part 2, Low Carb Router API: Router, Route, Link, Switch
  */
 
-export const Router = (props) => {
+export const Router = ({ hook, base, matcher, children }) => {
   const ref = useRef();
+  const { base: parentBase } = useRouter();
 
   // this little trick allows to avoid having unnecessary
   // calls to potentially expensive `buildRouter` method.
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
-  const value = ref.current || (ref.current = { v: buildRouter(props) });
+  const value =
+    ref.current ||
+    (ref.current = {
+      v: buildRouter({
+        hook,
+        base: parentBase ? parentBase + (base ?? "") : base,
+        matcher,
+      }),
+    });
 
   return h(RouterCtx.Provider, {
     value,
-    children: props.children,
+    children,
   });
 };
 
--- a/test/use-router.test.js
+++ b/test/use-router.test.js
@@ -59,3 +59,36 @@ it("shares one router instance between components", () => {
   ];
   expect(uniqRouters.length).toBe(1);
 });
+
+it("nests non-empty base paths", () => {
+  const { result } = renderHook(() => useRouter(), {
+    wrapper: (props) => (
+      <Router base="/base">
+        <Router base="/nested">{props.children}</Router>
+      </Router>
+    ),
+  });
+  expect(result.current.base).toBe("/base/nested");
+});
+
+it("handles empty parent base paths", () => {
+  const { result } = renderHook(() => useRouter(), {
+    wrapper: (props) => (
+      <Router>
+        <Router base="/nested">{props.children}</Router>
+      </Router>
+    ),
+  });
+  expect(result.current.base).toBe("/nested");
+});
+
+it("propages the base path", () => {
+  const { result } = renderHook(() => useRouter(), {
+    wrapper: (props) => (
+      <Router base="/base">
+        <Router>{props.children}</Router>
+      </Router>
+    ),
+  });
+  expect(result.current.base).toBe("/base");
+});


duarten:
ping?

jvdsande:
This is a breaking change though, in our app we use the fact that base is not inherited by default, and we use custom base-inheriting routers where necessary.

If this is implemented directly in Wouter's Router component, it should be put behind a prop, disabled by default, such as:

```
<Router base="/a">
  <Router base="/b" inheritBase>
    // Here base is "/a/b"
  </Router>     
  <Router base="/c">
    // Here base stays as "/c"
  </Router>
</Router>
```

HansBrende:
@jvdsande the example given in the docs is simply not acceptable for real-world use cases. (That being said, I am also not necessarily a fan of the implementation given in this PR.) To implement this correctly requires some hackery **way** beyond what you see in the toy docs example. Here's the hackery I've resorted to to get this working myself:

```typescript
function splitPath(pattern: Path, path: Path, matchParams: DefaultParams) {
    const matchKey = /:([^/:]+)\*$/.exec(pattern);
    if (!matchKey) return null;
    const rest = matchParams[matchKey[1]];
    if (rest == null || !path.endsWith("/" + rest)) return null;
    return [path.substring(0, path.length - rest.length - 1), rest];
}

export const Route = (props: RouteProps) => {
    const {path: pattern, children, component: Component} = props;

    const parentProps = useRouter();
    const [pathToMatch] = parentProps.hook(parentProps);

    const [matches, params] = (props as any).match as Match ?? parentProps.matcher(pattern ?? '', pathToMatch);
    if (!matches) {
        return null;
    }
    const [base, rest] = splitPath(pattern ?? '', pathToMatch, params!) ?? [pathToMatch, ''];
    return <RouteContext parent={parentProps} base={base}>
        <RemainingPathContext.Provider
            value={rest}
            children={Component ? <Component params={params!}/>
                : typeof children === "function" ? children(params!) : children}
        />
    </RouteContext>;
};

const RemainingPathContext = React.createContext<string>(null!);

const useCtxLocationHook = (ctx: RouterProps): [Path, (to: Path, options?: {replace?: boolean}) => void] => [
    React.useContext(RemainingPathContext),
    useNavigateFromRouterContext(ctx)
]

function RouteContext({parent, base, children}: {parent: RouterProps, base: string, children?: ReactNode}) {
    const router = Router({hook: useCtxLocationHook, base, children, matcher: parent.matcher})
    const v = router!.props.value.v;
    if (!v._parent) {
        Object.defineProperty(v, 'base', {get: () => `${v._parent.base}${v._relbase}`});
    }
    v._parent = parent;
    v._relbase = base;
    return router;
}

const useNavigateFromRouterContext = (ctx: RouterProps) => React.useCallback(
    (to: Path, options?: {replace?: boolean}) => {
        window.history[options?.replace ? "replaceState" : "pushState"](null, "",
            to[0] === "~" ? to.slice(1) : ctx.base + (to && !to.startsWith('/') && !ctx.base.endsWith('/') ? `/${to}` : to))
    }, [ctx]); // Do NOT depend on ctx.base! Unreliable, and also will cause this callback to un-memoize.

export const useNavigate = () => useNavigateFromRouterContext(useRouter());
```

We can now forget about `Router`s altogether and do:
```typescript
<Route path='/a/:rest*'> // <-- mounted under base ""
    <AComponent/>
</Route>

function AComponent() {
    const navigate = useNavigate(); // <-- navigates relative to base path "/a"
    return <Route path='/b/:rest*'> // <-- mounted under base "/a"
        <BComponent/>
    </Route>
}

function BComponent() {
    const navigate = useNavigate(); // <-- navigates relative to base path "/a/b"
}
```

Assuming https://github.com/molefrog/wouter/issues/230 is fixed at some point, I'll later simplify this hack to work with the path patterns `/a/*` and `/b/*`.

HansBrende:
Note: to make a change like this backwards-compatible and address @jvdsande's concerns, could simply relativize paths only when they have no leading slash. That's what react-router did. Definitely not a fan whatsoever of adding an extra flag... that will translate into a lot of extra boilerplate. Or, just release a new major version (my preference, since an absolute nested route makes zero sense unless it is prefixed with its parent's base path anyway).

jvdsande:
For what it's worth I was not saying that the example in the doc was production-ready, but that its mere existence showed that the behavior not being default was intended in the library, and not something that needed to be "fixed" :)

I agree with you that in real app you need something a bit more complex, and you can't really rely on the `key` hack used in the example since you want your subtree to update, not be nuked and replaced ^^"

That being said we got away with code way less complex that what you show (basically wrapping Router along a listener to its parent base, dynamically updating the base value stored in context).

In the end, the way to officially support this out-of-the-box inside Wouter needs to be @molefrog 's decision.

jvdsande:
On a side note: we are indeed using _absolutely nested routes_ and they actually make sense in our use case: microfrontends that handle their own subrouting, but have no idea in advance on which base route they will be mounted.

Wouter's current designed allowed us to develop each microfrontend as a standalone app using absolute routes as if it were mounted on /, and never having to worry about passing down the actual mount point to the subtree.

HansBrende:
@jvdsande I think one option here that would help both parties is to allow `Router` to accept an _optional_ parent router context value. If the parent router is supplied then `base` would resolve relative to the parent router's base. That would guarantee full backwards compatibility, and eliminate the ugliness of the hackery needed to get relative routing structures to work for people with that use-case.

HansBrende:
@jvdsande @duarten @molefrog I created an alternative [PR](#231) to this one that supports full flexibility & backwards compatibility while remaining super light-weight. Lemme know your thoughts!

duarten:
@HansBrende that's better indeed! Closing this PR in favour of #231. Thanks for taking the time to write it :)

PR #231, opened 2022-01-30
Make a more robust nested routing structure possible

### LINKED ISSUE #225: Base path inconsistencies

### LINKED ISSUE #257: Deeply nested routes
Following the [docs](https://github.com/molefrog/wouter/#are-relative-routes-and-links-supported) for nested routes, it seems to be hinted at that you can nest routes logically:

```
<Router base="/app">
	<NestedRoutes base="/dashboard">
		{/* the real url is /app/dashboard/users */}
		<Link to="/users" />
		<Route path="/users" />
	</NestedRoutes>
</Router>
 ```

However, when trying things out in the [demo sandbox](https://codesandbox.io/s/wouter-demo-nested-routes-ffd5h), this doesn't seem to be the case. Perhaps I'm doing something wrong?

### Steps to reproduce

1. Add a link to `/other` to the navigation list under `<Scope base='help'>`.
2. Add `<Scope base="/other"><p>Example text</p></Scope>` above `<Route path="/topics">`.

### Expected result:

`<p>Example text</p>` is rendered (like `/topics` and `/how-to`) at `/help/other`.

### Actual result:

`<p>Example text</p>` is not rendered at `/help/other`.

You can achieve a similar scenario by wrapping the main ` <div className="App">` with ` <Router base='/app'>`. This will render everything up to `/help` normally. There seems to be a bit of confusion as to how to achieve nested routes beyond the first level.

I'd appreciate any help regarding the matter! We're trying to transition away from **React Router** and its backward-facing development.

### PR DESCRIPTION
**Pull request by HansBrende**
Here's my basic idea. Fixes #225 and fixes #257 while remaining fully backwards compatible. Alternative to #226. For the nested router use-case, it will now be possible to do:

```typescript
function NestedRouter(props) {
    return <Router {...props} parent={useRouter()}/>
}
```

If this PR gets some traction, I can update type definitions as needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,8 +26,17 @@ const RouterCtx = createContext({});
 const buildRouter = ({
   hook = locationHook,
   base = "",
-  matcher = makeMatcher(),
-} = {}) => ({ hook, base, matcher });
+  matcher,
+  parent,
+} = {}) => ({
+  hook,
+  relbase: base,
+  parent,
+  matcher: matcher ?? parent?.matcher ?? makeMatcher(),
+  get base() {
+    return (this.parent?.base ?? "") + this.relbase;
+  },
+});
 
 export const useRouter = () => {
   const globalRef = useContext(RouterCtx);
@@ -67,6 +76,13 @@ export const Router = (props) => {
   // calls to potentially expensive `buildRouter` method.
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
   const value = ref.current || (ref.current = { v: buildRouter(props) });
+
+  // Use the "latest ref" pattern.
+  // https://epicreact.dev/the-latest-ref-pattern-in-react/
+  useLayoutEffect(() => {
+    value.v.relbase = props.base || "";
+    value.v.parent = props.parent;
+  });
 
   return h(RouterCtx.Provider, {
     value,
--- a/index.js
+++ b/index.js
@@ -30,11 +30,11 @@ const buildRouter = ({
   parent,
 } = {}) => ({
   hook,
-  relbase: base,
+  ownBase: base,
   parent,
   matcher: matcher ?? parent?.matcher ?? makeMatcher(),
   get base() {
-    return (this.parent?.base ?? "") + this.relbase;
+    return (this.parent?.base ?? "") + this.ownBase;
   },
 });
 
@@ -80,7 +80,7 @@ export const Router = (props) => {
   // Use the "latest ref" pattern.
   // https://epicreact.dev/the-latest-ref-pattern-in-react/
   useLayoutEffect(() => {
-    value.v.relbase = props.base || "";
+    value.v.ownBase = props.base || "";
     value.v.parent = props.parent;
   });
 


HansBrende:
FYI, this PR will also allow some really cool stuff like a _stable_ `useNavigate` function relative to your current router context. E.g.:

```typescript
const useNavigateFromRouterContext = (ctx: RouterProps) => React.useCallback(
    (to: Path, options?: {replace?: boolean}) => {
        window.history[options?.replace ? "replaceState" : "pushState"](null, "",
            to[0] === "~" ? to.slice(1) : 
            ctx.base + (to && !to.startsWith('/') && !ctx.base.endsWith('/') ? `/${to}` : to))
    }, [ctx]);

export const useNavigate = () => useNavigateFromRouterContext(useRouter());
```

index.js
```diff
@@ -26,8 +26,17 @@ const RouterCtx = createContext({});
 const buildRouter = ({
   hook = locationHook,
   base = "",
-  matcher = makeMatcher(),
-} = {}) => ({ hook, base, matcher });
+  matcher,
+  parent,
+} = {}) => ({
+  hook,
+  relbase: base,
```
jvdsande, line 33: I'm not a big fan of the naming `relbase` here. Is it a contraction of `relative` + `base`? In this case I'd at least make it camelcase (`relBase`). I'd even go for `relativeBase` but that adds a few bytes which is a shame for so little gain.

jvdsande (commented):
I indeed prefer this approach to the one in #226. It's kind of what I asked with a flag, but cleaner and more powerful since we can play with _which_ parent we give to our Router.

index.js
```diff
@@ -68,6 +77,13 @@ export const Router = (props) => {
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
   const value = ref.current || (ref.current = { v: buildRouter(props) });
 
+  // Use the "latest ref" pattern.
+  // https://epicreact.dev/the-latest-ref-pattern-in-react/
+  useLayoutEffect(() => {
+    value.v.relbase = props.base || "";
+    value.v.parent = props.parent;
+  });
```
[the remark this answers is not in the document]
HansBrende: @jvdsande yes: your approach definitely won't work in concurrent mode with React 18! Because the order of render operations could be unpredictable. Your ref might contain values from a stale and/or cancelled render operation, making it unreliable! This is why side effects must never happen in render. (Lazy initialization is the big exception, because it's not really a side effect.)

index.js
HansBrende: I could do camelcase for sure... yes, "relative" + "base" was what I was going for.

index.js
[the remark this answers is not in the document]
jvdsande: Ah yes, definitely. Did not think of that :)

index.js
HansBrende: Pondered this question some more... I think I will switch to `ownBase`! Short and sweet and completely obvious semantics.

HansBrende:
@molefrog any thoughts on this PR? I can update the type definitions and stuff if you like my changes.

Scott-Fischer:
I was about to begrudgingly switch to react-router but this PR makes me think I should wait! I'll be keeping an eye on this PR because it's something that I really need in my current project.

molefrog:
Closing in favour of #265

Issue #237, opened 2022-03-11
Prevent navigation on Link component if event was cancelled


Tomas2D:
Due to calling `onEvent` (passed from props) after doing navigation, one cannot prevent cancelling the navigation process.

I propose to call `onClick` before `navRef.current();` and then check if event was cancelled via `e.isDefaultPrevented()`. If so, then do not call `navRef.current()` action. What do you think?

https://github.com/molefrog/wouter/blob/cba34af73d01389be18d0ed275611336946c18fe/index.js#L98-L118

Tomas2D:
I able to make a PR if you do not have a time, that is not a problem. @molefrog

molefrog:
That would make a lot of sense, is `event.isDefaultPrevented` part of the web standard? To my shame, I haven't used this method. 

Also, PRs are highly welcomed!

Tomas2D:
It looks like that `defaultPrevented` property is, also with high [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented#browser_compatibility) as can be seen here. 

This simple [https://codesandbox.io/s/divine-bash-lhoxd7?file=/src/App.js](codesandbox) demonstrate the funcionality.

I will create an PR.

molefrog:
Landed in https://github.com/molefrog/wouter/releases/tag/v2.8.0 🎊

PR #239, opened 2022-03-18
Prevent navigation when event was cancelled

Tomas2D:
Issue described in #237

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,9 +108,11 @@ export const Link = (props) => {
       )
         return;
 
-      event.preventDefault();
-      navRef.current();
       onClick && onClick(event);
+      if (!event.defaultPrevented) {
+        event.preventDefault();
+        navRef.current();
+      }
     },
     // navRef is a ref so it never changes
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


Tomas2D:
ping @omgovich

Tomas2D:
ping @molefrog

index.js
```diff
@@ -108,9 +108,13 @@ export const Link = (props) => {
       )
         return;
 
+      onClick && onClick(event);
+      if (event.defaultPrevented) {
+        return;
+      }
```
molefrog, line 114: Maybe we save a few bytes by rewriting it as:

```suggestion
      if (!event.defaultPrevented) {
        event.preventDefault();
        navRef.current()
      }
```

?

index.js
```diff
@@ -108,9 +108,13 @@ export const Link = (props) => {
...
+
       event.preventDefault();
       navRef.current();
-      onClick && onClick(event);
```
molefrog, line 113: I'm trying to think about any possible use case when the order of `onClick` and `navRef()` would matter. It doesn't seem so at the first glance, but maybe @Tomas2D you have some thoughts?

molefrog (approved):
This looks good, thank you @Tomas2D!

I've left some minor comments and suggestions.

index.js
Tomas2D: Make sense, why not.

molefrog (approved):

molefrog:
Thanks for your contribution @Tomas2D!

I'm going to work on releasing the stable 2.8.0 and mention you in the changelog when it's ready.

PR #251, opened 2022-07-08
Remove the unnecessary search state

HelKyle:


cbbfcd (approved):
LGTM

use-location.js
```diff
@@ -9,11 +9,9 @@ const eventReplaceState = "replaceState";
 export const events = [eventPopstate, eventPushState, eventReplaceState];
 
 export default ({ base = "" } = {}) => {
-  const [{ path, search }, update] = useState(() => ({
-    path: currentPathname(base),
-    search: location.search,
-  })); // @see https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
-  const prevHash = useRef(path + search);
+  const [{ path }, update] = useState(() => ({ path: currentPathname(base) }));
```
molefrog, line 12:
```suggestion
  const [path, update] = useState(() => currentPathname(base));
```

Would that work?

molefrog: No worries, I can fix it in the main branch.

molefrog: Turns out I was wrong here, if we change the state type to string, React won't trigger a re-render (because it's a primitive value and it doesn't change).

HansBrende:
@HelKyle @cbbfcd @molefrog not sure I understand the point of this PR. Did this fix anything? Doesn't look like anything has changed here semantically (except maybe that the codebase has now become more unstable due to decoupling the points at which location.pathname & location.search are read). No corresponding issue or description... can someone explain this to me?

location.search is indeed unused, but this PR did not remove it as suggested by the title.

molefrog:
> @HelKyle @cbbfcd @molefrog not sure I understand the point of this PR. Did this fix anything? Doesn't look like anything has changed here semantically (except maybe that the codebase has now become more unstable due to decoupling the points at which location.pathname & location.search are read). No corresponding issue or description... can someone explain this to me?
> 
> location.search is indeed unused, but this PR did not remove it as suggested by the title.

I kinda agree, I might have been too light-hearted here. 

However, wouldn't you think that keeping search in the state wasn't necessary? I mean, it was never used and we left it there so that we could implement search string feature someday. Unfortunately, we haven't done this yet (I believe it requires proper design first, and I didn't really have time to work on it yet).

HansBrende:
@molefrog technically it is still in the state since it is included in the ref (which is just state that never triggers a render). But whether or not it is in the state coming from `useState` has no effect on rendering in this case (since it is the changing object containing those properties that triggers the render, not the properties themselves), therefore the two are semantically equivalent with the exception that different parts of `location` are now read in different places: an... interesting... choice... 😄 

In any case, I'd highly recommend replacing everything here with `useSyncExternalStore` ASAP. I don't trust this part of the code at all and relying on it last year after some flaky stuff happened (occasionally not updating when it should). After I replaced it in my own codebase with `useSyncExternalStore` the flakiness disappeared.

One idea: add a dependency on the `useSyncExternalStore` shim package for versions < 3.x, then remove that dependency in 3.x.

HansBrende:
@molefrog For the record, here's how I've implemented this in my codebase:

```typescript
const events = ["popstate", "pushState", "replaceState", "hashchange"];
export function subscribeToLocation(callback: () => void) {
    for (const event of events) {
        window.addEventListener(event, callback);
    }
    return () => {
        for (const event of events) {
            window.removeEventListener(event, callback);
        }
    };
}

const getCurrentPath = () => window.location.pathname;

const getCurrentSearch = () => window.location.search;

export function usePathname() {
    return useSyncExternalStore(subscribeToLocation, getCurrentPath);
}

export function useSearchParams() {
    const search = useSyncExternalStore(subscribeToLocation, getCurrentSearch);
    return new URLSearchParams(search);
}
```

I separated these out so that renders aren't triggered unnecessarily (e.g., components that only rely on pathname don't rerender if only the search has changed).

Existing behavior could be replicated with something along the lines of:
```
export function useLocation(base) {
    useSearchParams(); // trigger pointless re-render on search change just for backwards compatibility 
    return [
        useSyncExternalStore(subscribeToLocation, () => calculatePathFromBase(getCurrentPath(), base)),
        useCallback(() => ...existing logic...)
    ]
}
```

PR #252, opened 2022-07-08
Pass buildRouter as a function to useState

HelKyle:


molefrog (approved):
Hi! Sorry that it took me so long to review this, I've been on a long break from doing open-source. This look good, thank you!

jorrit:
Just curious, why is useState without updates better than useRef?

HansBrende:
@molefrog @HelKyle @jorrit I am also curious about this... why was this change made? Not seeing any rationale to explain the change. Can someone please explain it to me?

molefrog:
> @molefrog @HelKyle @jorrit I am also curious about this... why was this change made? Not seeing any rationale to explain the change. Can someone please explain it to me?

There really isn't any difference in performance, probably the only benefit is that it is shorter than `useRef`. Plus, it is now (I believe they rewrote this chapter from the docs, it used to be just `useRef` approach) the recommended way of doing memoized initialisation. See the link from the comment: https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily

So it is mostly a stylistic choice, I'm good with both versions.

Btw, regarding the comment, I think we could make it shorter, e.g. `// initialize the router only once`, don't think it requires so much attention.

HansBrende:
@molefrog gotcha, I see. Don't think it is ***the*** recommended way necessarily (since the description on how to lazy-init refs is still present in that same section, under "**You might also occasionally want to avoid re-creating the useRef() initial value**.") 

But agreed that it is purely a stylistic choice without much of a performance impact, was just curious what the rationale was. Thanks!

Issue #256, opened 2022-07-28
[feature] Wrap Link with forwardRef


avionbg:
There are cases where we need a working ref to rendered link dom node, it would really be helpful if this gets into the code.

cbbfcd:
can you tell me what the scene is?

avionbg:
Most typical example would be MUI <Button component={Link} /> - which fails due to the missing ref.

trymoto:
Same here, would be useful with MUI

trymoto:
@cbbfcd I can go ahead and create a pr if it's ok, should be pretty straightforward.

molefrog:
This has been implemented and released in https://github.com/molefrog/wouter/releases/tag/v2.8.0

Thanks @trymoto for your help!

PR #259, opened 2022-10-15
Wrap Link with forwardRef

trymoto:
Hi!
This relates to [#256](https://github.com/molefrog/wouter/issues/256)

Wrapping components with `forwardRef` is a common way component libraries expose underlying DOM to library users.
One particular example where it might be useful is passing `Link` down as a component to use in various navigation elements, e.g. MUI's `Buttons` and `Tabs` [See related doc section](https://mui.com/material-ui/guides/routing/#component-prop). Currently one'd write their own wrapper with same functionality as `Link`, but i thought it would be better to have wrapper out of the box. There aren't drawbacks i could think of, if you know about one, please, leave a comment.

Thanks.

### DIFF
--- a/test/link.test.js
+++ b/test/link.test.js
@@ -3,6 +3,20 @@ import { render, cleanup, fireEvent } from "@testing-library/react";
 
 import { Router, Link } from "../index.js";
 
+const ParentComponent = (props) => {
+  const ref = React.createRef();
+
+  React.useEffect(() => {
+    ref.current.innerHTML = "Tested";
+  }, [ref])
+
+  return (
+    <Link href="/about" ref={ref}>
+      {props.children}
+    </Link>
+  );
+}
+
 afterEach(cleanup);
 
 it("renders a link with proper attributes", () => {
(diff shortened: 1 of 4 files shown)


cbbfcd (approved):
LGTM

trymoto:
kindly pinging @molefrog

trymoto:
It's my understanding that **molefrog** is unavailable, is there anyone else empowered to merge and release **wouter**?

molefrog:
Hi, @molefrog is here. Sorry it's been a crazy year for me. I'll take a look next week 🖖

test/link.test.js
```diff
@@ -3,6 +3,20 @@ import { render, cleanup, fireEvent } from "@testing-library/react";
 
 import { Router, Link } from "../index.js";
 
+const ParentComponent = (props) => {
```
molefrog, line 6: Can we rename this to something like `LinkWithForwardedRef` or similar just to avoid confusion with the rest of the tests?

molefrog (changes requested):
Thanks, this actually looks good!

One thing though is that we don't require from our Preact users to have `preact/compat` installed (some people prefer to keep it as small as possible). We'll have to find some workaround, otherwise we risk breaking compatibility with previous versions of Preact lib.

test/link.test.js
trymoto: Sure

trymoto:
> Thanks, this actually looks good!
> 
> One thing though is that we don't require from our Preact users to have `preact/compat` installed (some people prefer to keep it as small as possible). We'll have to find some workaround, otherwise we risk breaking compatibility with previous versions of Preact lib.

Yep seems reasonable, i'll see what we can do.

molefrog:
> > Thanks, this actually looks good!
> > One thing though is that we don't require from our Preact users to have `preact/compat` installed (some people prefer to keep it as small as possible). We'll have to find some workaround, otherwise we risk breaking compatibility with previous versions of Preact lib.
> 
> Yep seems reasonable, i'll see what we can do.

I did some brief digging today and it looks like what Preact doing to support `forwardRef` isn't super sophisticated, however is does patch the rendering pipeline https://github.com/preactjs/preact/blob/master/compat/src/forwardRef.js#L5

So I think we have two options here:
1. For Preact, provide `forwardRef` implementation that simply does nothing and just forwards the original component
2. Or, introduce another prop called `innerRef` (inspired by [this](https://styled-components.com/docs/api#deprecated-innerref-prop)) which can be used in environments where `forwardRef` isn't available (we can support the prop in React as well, it can be just an alias to `ref`).

trymoto:
Added forwardRef stub for now

molefrog (approved):
Looks great!

molefrog:
Released in https://github.com/molefrog/wouter/releases/tag/v2.8.0

PR #262, opened 2022-11-03
TS: Parameter values are optional when can't be inferred 

### LINKED ISSUE #214: DefaultParams type doesn't count with optional parameters
Wouter export types:

```typescript
export interface DefaultParams {
  [paramName: string]: string;
}
export type Params<T extends DefaultParams = DefaultParams> = T;
```

Let's use dynamic segment in the path, eg:

```
<Route path="/:foo/:bar?">
  {(params) => {
    
  }}
</Route>;
```

...and match the route against path `/hello`. Then matcher returns params as following: `{ foo: hello, bar: undefined }`. But those params doesn't comply with `DefaultParams` definition. So I think it should be:

```typescript
export interface DefaultParams {
  [paramName: string]: string | undefined;
}
```

### LINKED ISSUE #229: Typescript Improvements possible (in v2.8.0-alpha.2)
First of all... **bravo** for making this library. I became a little frustrated with react-router last weekend and started redesigning everything in it... only to come here and find this library, which I find has **very similar ideas** to my homebrewed stuff! There are just a couple areas in the typings where my library caught some edge cases and yours does not... so I will share some happy paths and sad paths below (The quoted errors are the errors received from *my* homebaked implementation. Anything with a red ❌ is where the behavior is unexpectedly different). Hopefully this screenshot gives you some ideas:
Screen Shot 2022-01-24 at 7 42 11 PM

molefrog:
Partially addresses #229
Closes #214

- When route parameters can't be inferred at a compile time, they become optional now. Routes may contain optional segments (e.g. `/:rest*`) so some parameters could be `undefined`
- Parameter values are now `readonly`

### DIFF
--- a/preact/types/matcher.d.ts
+++ b/preact/types/matcher.d.ts
@@ -5,18 +5,14 @@
 import { Path } from "./use-location";
 
 export interface DefaultParams {
-  [paramName: string]: string;
+  readonly [paramName: string]: string | undefined;
 }
+
 export type Params<T extends DefaultParams = DefaultParams> = T;
 
-export type MatchWithParams<T extends DefaultParams = DefaultParams> = [
-  true,
-  Params<T>
-];
+export type MatchWithParams<T extends DefaultParams = DefaultParams> = [true, Params<T>];
 export type NoMatch = [false, null];
-export type Match<T extends DefaultParams = DefaultParams> =
-  | MatchWithParams<T>
-  | NoMatch;
+export type Match<T extends DefaultParams = DefaultParams> = MatchWithParams<T> | NoMatch;
 
 export type MatcherFn = (pattern: Path, path: Path) => Match;
 
--- a/preact/types/ts4.1/index.d.ts
+++ b/preact/types/ts4.1/index.d.ts
@@ -1,13 +1,7 @@
 // Minimum TypeScript Version: 4.1
 // tslint:disable:no-unnecessary-generics
 
-import {
-  JSX,
-  FunctionComponent,
-  ComponentType,
-  ComponentChildren,
-  VNode,
-} from "preact";
+import { JSX, FunctionComponent, ComponentType, ComponentChildren, VNode } from "preact";
 
 import {
   Path,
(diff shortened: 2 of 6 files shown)


molefrog:
Hey @HansBrende, I hope this is still relevant. Once again, sorry for my late reply. I've updated type definitions based on your suggestions, do you mind taking a look?

Thanks!

HansBrende:
@molefrog sure thing, thanks a lot! Will take a look tomorrow. Also since you now seem to be back in business, I have a PR open which I've been waiting for some kind of resolution for (would love to strip the hacks out of my current code) if you'd be so kind as to take a look: https://github.com/molefrog/wouter/pull/231

HansBrende:
@molefrog thanks! I've taken a look and it appears to be working as expected now. (Note however that I've only tested the typescript 4.1 and standard React flavors, I have not tested on Preact or out-of-date typescript versions).

Also note [my comment](https://github.com/molefrog/wouter/issues/229#issuecomment-1303825323) describing how the second half of the issue I raised is fixed in React 18, but still present before React 18.

Side note: since you now seem to be back in business, I have a PR open which I've been waiting for some kind of resolution for (would love to replace the hacks in my current codebase with this simple fix) if you'd be so kind as to take a look: https://github.com/molefrog/wouter/pull/231

HansBrende (approved):

molefrog:
> @molefrog thanks! I've taken a look and it appears to be working as expected now. (Note however that I've only tested the typescript 4.1 and standard React flavors, I have not tested on Preact or out-of-date typescript versions).
> 
> Also note [my comment](https://github.com/molefrog/wouter/issues/229#issuecomment-1303825323) describing how the second half of the issue I raised is fixed in the React 18 typings, but still present before React 18.
> 

Thank you for the review. 

The second half of the issue is to be addresses in the next PR, I've checked it and it is a valid point.

> Side note: since you now seem to be back in business, I have a PR open which I've been waiting for some kind of resolution for (would love to replace the hacks in my current codebase with this simple fix) if you'd be so kind as to take a look: #231

Yep, in my to-do list.

PR #263, opened 2022-11-05
React < 18: Fix incorrect `ReactNode` declaration

### LINKED ISSUE #229: Typescript Improvements possible (in v2.8.0-alpha.2)

molefrog:
Closes #229

This PR declares more strict `ReactNode` that excludes `{}`. Only relevant to React.
It also fixes some incorrect test cases.

### DIFF
--- a/types/ts3.9.4/index.d.ts
+++ b/types/ts3.9.4/index.d.ts
@@ -6,7 +6,8 @@ import {
   PropsWithChildren,
   ComponentType,
   ReactElement,
-  ReactNode,
+  ReactChild,
+  ReactPortal,
 } from "react";
 
 import {
@@ -22,6 +23,11 @@ import { DefaultParams, Params, Match, MatcherFn } from "../matcher";
 // re-export types from these modules
 export * from "../matcher";
 export * from "../use-location";
+
+// React <18 only: fixes incorrect `ReactNode` declaration that had `{}` in the union.
+// This issue has been fixed in React 18 type declaration.
+// https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210
+type ReactNode = ReactChild | ReactNode[] | ReactPortal | boolean | null | undefined;
 
 /*
  * Components: <Route />
--- a/types/ts3.9.4/index.d.ts
+++ b/types/ts3.9.4/index.d.ts
@@ -27,7 +27,7 @@ export * from "../use-location";
 // React <18 only: fixes incorrect `ReactNode` declaration that had `{}` in the union.
 // This issue has been fixed in React 18 type declaration.
 // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210
-type ReactNode = ReactChild | ReactNode[] | ReactPortal | boolean | null | undefined;
+type ReactNode = ReactChild | Iterable<ReactNode> | ReactPortal | boolean | null | undefined;
 
 /*
  * Components: <Route />
--- a/types/ts4.1/type-specs.tsx
+++ b/types/ts4.1/type-specs.tsx
@@ -59,7 +59,7 @@ const invalidParamsWithGeneric: Params<{ id: number }> = { id: 13 }; // $ExpectE
   This is a <b>mixed</b> content
 </Route>;
 
-<Route path="/users/:id">{(params: Params): React.ReactNode => `User id: ${params.id}`}</Route>;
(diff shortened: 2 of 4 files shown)


types/ts4.1/type-specs.tsx
```diff
@@ -77,6 +77,8 @@ const invalidParamsWithGeneric: Params<{ id: number }> = { id: 13 }; // $ExpectE
   {({ rest }) => {
     const fn = (a: string) => "noop";
     fn(rest); // $ExpectError
+
+    return <></>;
```
molefrog, line 81: The previous definition was incorrect, but thanks to the fix I was able to catch it.

types/ts4.1/type-specs.tsx
```diff
@@ -59,7 +59,7 @@ const invalidParamsWithGeneric: Params<{ id: number }> = { id: 13 }; // $ExpectE
   This is a <b>mixed</b> content
 </Route>;
 
-<Route path="/users/:id">{(params: Params): React.ReactNode => `User id: ${params.id}`}</Route>;
+<Route path="/users/:id">{(params: Params) => `User id: ${params.id}`}</Route>;
```
molefrog, line 62: This one is interesting. Because we use React 17 for testing the types, the two `ReactNode` versions were conflicting with each other. 

This could lead to some cases where a user decides to define their function explicitly returning `ReactNode` and they might run into this issue. But I suppose since it was a bug they, unfortunately, would have to fix it themselves.

types/ts3.9.4/index.d.ts
```diff
@@ -23,6 +24,11 @@ import { DefaultParams, Params, Match, MatcherFn } from "../matcher";
 export * from "../matcher";
 export * from "../use-location";
 
+// React <18 only: fixes incorrect `ReactNode` declaration that had `{}` in the union.
```
molefrog, line 27: Side thought: honestly, supporting two versions of TS becomes a bit of a burden. I wonder if we could drop support for 3.9.4 in the future major version. Would that be too brutal? Imo, it could, but there is an alternative – can't upgrade to TS 4, well you would have to stay on wouter@2, sorry.

molefrog:
Hi @HansBrende, please see the solution provided.

types/ts3.9.4/index.d.ts
HansBrende: 🤷‍♂️  I have no idea. Personally, I don't use TS < 4 and never plan to. But others may have a different opinion on this 😄

types/ts3.9.4/index.d.ts
```diff
@@ -23,6 +24,11 @@ import { DefaultParams, Params, Match, MatcherFn } from "../matcher";
...
+// This issue has been fixed in React 18 type declaration.
+// https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210
+type ReactNode = ReactChild | ReactNode[] | ReactPortal | boolean | null | undefined;
```
HansBrende, line 30: Looks good. Only other thing I will suggest is to change
`ReactNode[]` to `Iterable<ReactNode>`. This was a change in DT that I overlooked. See details here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56623

Also worth noting, react has allowed Iterables prior to 18, it was always just incorrectly typed in DT (a problem which was masked by the inclusion of `{}`). https://github.com/facebook/react/blob/v17.0.2/packages/shared/ReactTypes.js#L20

HansBrende (changes requested):

types/ts3.9.4/index.d.ts
molefrog: Good one! Fixed in https://github.com/molefrog/wouter/pull/263/commits/a3c480e87a9e3e831a726765853fe130c1241021

HansBrende (approved):
Tested on React + TS4.1, looks good!

molefrog:
Thank you @HansBrende, I have just published `2.8.1`.

PR #265, opened 2022-11-15
Support for `parent` prop in Router

molefrog:
It's an alternative solution to #231.
Notable differences:
1. Because `<Router />` is immutable (it only gets created once and doesn't react on props change) we can simply make a copy of the parent router, it does not have to be dynamic.
2. Changed the API, there is no need now to write `parent={useRouter()}` which IMO could be not obvious for some users. Instead, by adding a `nested` prop, we the the router to inherit the base path. I even think that we can enable this by default in future releases.
3. Simplified the context it passed an object instead of a reference. This might sound like a performance downgrade, but in reality because Router is immutable it never changes between renders. Hence, context update won't be triggered.

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,7 +81,7 @@ export const Router = ({
     router.base = proto.base + base;
 
     // parent router reference, `undefined` when router is top-level
-    router.parent = proto === defaultRouter ? proto : undefined;
+    router.parent = proto === defaultRouter ? undefined : proto;
 
     return router;
   };
--- a/index.js
+++ b/index.js
@@ -61,7 +61,13 @@ const useNavigate = (options) => {
  * Part 2, Low Carb Router API: Router, Route, Link, Switch
  */
 
-export const Router = ({ hook, matcher, base = "", nested = false, children }) => {
+export const Router = ({
+  hook,
+  matcher,
+  base = "",
+  nested = false,
+  children,
+}) => {
   const parent = useRouter();
 
   // nested `<Router />` has the scope of its closest parent router (base path is prepended)
@@ -115,7 +121,13 @@ export const Link = forwardRef((props, ref) => {
     (event) => {
       // ignores the navigation when clicked using right mouse button or
       // by holding a special modifier key: ctrl, command, win, alt, shift
-      if (event.ctrlKey || event.metaKey || event.altKey || event.shiftKey || event.button !== 0)
+      if (
+        event.ctrlKey ||
+        event.metaKey ||
+        event.altKey ||
+        event.shiftKey ||
+        event.button !== 0
+      )
         return;
 
       onClick && onClick(event);
@@ -146,7 +158,9 @@ const flattenChildren = (children) => {
   return Array.isArray(children)
     ? [].concat(
         ...children.map((c) =>
-          c && c.type === Fragment ? flattenChildren(c.props.children) : flattenChildren(c)
+          c && c.type === Fragment
+            ? flattenChildren(c.props.children)
+            : flattenChildren(c)
         )
       )
     : [children];
--- a/index.js
+++ b/index.js
@@ -61,7 +61,13 @@ const useNavigate = (options) => {
  * Part 2, Low Carb Router API: Router, Route, Link, Switch
  */
 
-export const Router = ({ hook, matcher, base = "", nested = false, children }) => {
+export const Router = ({
+  hook,
+  matcher,
+  base = "",
+  nested = false,
+  children,
+}) => {
   const parent = useRouter();
 
   // nested `<Router />` has the scope of its closest parent router (base path is prepended)
@@ -110,7 +116,13 @@ export const Link = forwardRef((props, ref) => {
     (event) => {
       // ignores the navigation when clicked using right mouse button or
       // by holding a special modifier key: ctrl, command, win, alt, shift
-      if (event.ctrlKey || event.metaKey || event.altKey || event.shiftKey || event.button !== 0)
+      if (
+        event.ctrlKey ||
+        event.metaKey ||
+        event.altKey ||
+        event.shiftKey ||
+        event.button !== 0
+      )
         return;
 
       onClick && onClick(event);
@@ -141,7 +153,9 @@ const flattenChildren = (children) => {
   return Array.isArray(children)
     ? [].concat(
         ...children.map((c) =>
-          c && c.type === Fragment ? flattenChildren(c.props.children) : flattenChildren(c)
+          c && c.type === Fragment
+            ? flattenChildren(c.props.children)
+            : flattenChildren(c)
         )
       )
     : [children];
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 import locationHook from "./use-location.js";
-import makeMatcher from "./matcher.js";
+import matcherWithCache from "./matcher.js";
 
 import {
   useRef,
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -62,15 +61,33 @@ const useNavigate = (options) => {
...
+  // side effects (e.g. you can't just hot swap the hook, it'll break the routing)
+  //
+  // we use `useState` here, but it only catches the first render and never changes.
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
-  const [value] = useState(() => ({ v: buildRouter(props) }));
+  const [value] = useState(() => ({
+    hook: hook || proto.hook,
+    matcher: matcher || proto.matcher,
+    base: proto.base + base,
+  }));
```
HansBrende, line 86: Looks good, except we **_absolutely do_** still need a layout effect here to update props (for **AT LEAST** `base`, preferably `matcher` as well).
Consider the following JSX:

```typescript
return app1Flag ? 
    <Router base="/app1">{children}</Router> 
    : 
    <Router base="/app2">{children}</Router>;
```
This code will fail dramatically if `base` is not updated within your ref (because React will simply change the props of the current component). My own use-cases would not work so I would still have to resort to hacks. In short: _we cannot assume that routers are immutable_. Now, one thing we can do is assume that the `hook` is immutable and simply not update the hook (a most-likely valid assumption). But no need to throw the baby out with the bathwater, because we can expect `base` to change frequently, depending on what part the app is currently being rendered.

Here's my recommended addition:

```typescript
useLayoutEffect(() => {
    value.matcher = matcher || proto.matcher;
    value.base = proto.base + base;
});
```

In addition, it would certainly be nice to have a reference to the parent router (e.g., if we want to navigate from the context of the _parent's_ base path in some custom operation.)

So I would further recommend the following change:

```typescript
const [value] = useState(() => ({
  hook: hook || proto.hook,
  matcher: matcher || proto.matcher,
  base: proto.base + base,
  parent: proto // <-- inserted here (note: `parent` for the default router should still be undefined, so we can detect when we are at the top).
}));
useLayoutEffect(() => {
    value.matcher = matcher || proto.matcher;
    value.base = proto.base + base;
    value.parent = proto;
});
```

HansBrende (changes requested):

index.js
```diff
@@ -62,15 +61,33 @@ const useNavigate = (options) => {
...
+  matcher,
+  base = "",
+  nested = false,
+  children,
+}) => {
+  const parent = useRouter();
+
+  // nested `<Router />` has the scope of its closest parent router (base path is prepended)
+  // Routers are not nested by default, but this might change in future versions
+  const proto = nested ? parent : defaultRouter;
```
HansBrende, line 75: This change is ok... however... it will lead to inefficiencies down the road. Because you are going to want to change `<Route>` at some point to support nested routes (you can calculate `base` **automatically** from the path regexp!!!)... in which case, you will already have the parent router and can supply it directly to the nested `<Router>` without having to call `useContext()` a second time redundantly. In addition, for non-nested routers, you are now adding an extra hook call for no reason.

My recommendation would be to supply `parent` instead of a `nested` flag. This makes the routers more powerful. And then, turn your focus to allowing users to not have to use Routers at all, by instead allowing nested `Route` to work and auto-calculate the base for the nested router beneath them. ("nested" would be a good flag for Routes instead of Routers!)

index.js
molefrog: You have a point here, I'll update this part.

index.js
molefrog: > This change is ok... however... it will lead to inefficiencies down the road. Because you are going to want to change `<Route>` at some point to support nested routes (you can calculate `base` **automatically** from the path regexp!!!)... in which case, you will already have the parent router and can supply it directly to the nested `<Router>` without having to call `useContext()` a second time redundantly. In addition, for non-nested routers, you are now adding an extra hook call for no reason.
> 
> My recommendation would be to supply `parent` instead of a `nested` flag. This makes the routers more powerful. And then, turn your focus to allowing users to not have to use Routers at all, by instead allowing nested `Route` to work and auto-calculate the base for the nested router beneath them. ("nested" would be a good flag for Routes instead of Routers!)

Not sure if I follow, could you please provide some example on how router and nested routes could be used together?

HansBrende: @molefrog Sure, I will put some examples together from my own codebase. Will get back to you soon.

index.js
```diff
@@ -62,15 +61,28 @@ const useNavigate = (options) => {
...
+    router.hook = hook || proto.hook;
+    router.matcher = matcher || proto.matcher;
+    router.base = proto.base + base;
+  };
+
+  // we use `useState` here, but it only catches the first render and never changes.
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
-  const [value] = useState(() => ({ v: buildRouter(props) }));
+  const [value] = useState(() => ({})); // create the object once...
+  updateRouter(value); // then update it on each render
```
HansBrende, line 81: @molefrog this will work on React 17, but React 18 uses concurrent rendering so we really need this inside a `useLayoutEffect()` to guarantee the correct ordering. Otherwise, these properties may not be updated to reflect the current render, as renders may occur "out of order", or be cancelled. I mentioned this also in my PR in response to a (now-resolved) comment asking the same question.

HansBrende (changes requested):

index.js
molefrog, line 81: @HansBrende here is my first take, no parent yet, but we'll get to that later.
Initially this code was:

```js
// updates the current router with the props passed down to the component
const updateRouter = (router) => {
  router.hook = hook || proto.hook;
  router.matcher = matcher || proto.matcher;
  router.base = proto.base + base;

  return router;
}; 

const [value] = useState(() => updateRouter({})); // create the object once...
useLayoutEffect(() => updateRouter(value)); // then update it on each render
```

However I was receiving `useLayoutEffect` warning when testing it in SSR:
image

molefrog: Any ideas on how we can fix this?

HansBrende: @molefrog yes, actually I do. I've seen several big codebases follow a certain pattern to deal with this problem, one moment while I look it up.

HansBrende: @molefrog this is how react-redux fixes this problem:

https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.ts

They define a `useIsomorphicLayoutEffect` that resolves to `useLayoutEffect` on the client, and `useEffect` on the server.

index.js
```diff
@@ -73,12 +73,16 @@ export const Router = ({ hook, matcher, base = "", nested = false, children }) =
     router.hook = hook || proto.hook;
     router.matcher = matcher || proto.matcher;
     router.base = proto.base + base;
+
+    return router;
   };
```
HansBrende, line 84: Can slightly improve efficiency here by pulling `updateRouter` out of the `Router` function.

E.g.
```typescript
const updateRouter = (router, hook, matcher, proto) => {...}

export const Router = (...) => {
  ...
  const [value] = useState(() => updateRouter({}, hook, matcher, proto));
  useIsomorphicLayoutEffect(() => {
      updateRouter(value, hook, matcher, proto);
  });
}

Comment (molefrog) [reply index.js:84]: Yep, but it kinda forces us to perform destructuring. I agree that there is a performance benefit, but at the same time we're loosing some points in stylistics terms. Plus, creating anonymous funcs in render [should be relatively cheap](https://reactjs.org/docs/hooks-faq.html#are-hooks-slow-because-of-creating-functions-in-render)

File: index.js [hunk 5]
```diff
@@ -62,15 +61,41 @@ const useNavigate = (options) => {
...
+  const proto = nested ? parent : defaultRouter;
+
+  // updates the current router with the props passed down to the component
+  const updateRouter = (router) => {
+    router.hook = hook || proto.hook;
+    router.matcher = matcher || proto.matcher;
+    router.base = proto.base + base;
+
+    // parent router reference, `undefined` when router is top-level
+    router.parent = proto === defaultRouter ? proto : undefined;
```
molefrog, line 84: @HansBrende FYI the way parent is assigned.

molefrog: Whoops, it should be other way around!

HansBrende: @molefrog yes, flipped 😂 . Also please see my comment here: https://github.com/molefrog/wouter/pull/265#discussion_r1023228882 where I flesh out your requested example. Hopefully that makes sense?

molefrog: Done in f5fd814cc99ba9bb7798f8dc12ec3ebe51db32da, also added some assertions.

index.js
molefrog: Thanks, I'll review the code and get back to you tomorrow.

molefrog: See https://github.com/molefrog/wouter/pull/265/commits/5963014ed7ce69a31240b43d11d71fd188903aae

HansBrende (approved):
Looks awesome! 🎉

molefrog:
Great! Thank you for your help @HansBrende.
I was thinking that we can add support for `useSyncExternalStore` next and include it in the same release. Would you want to work on that?

HansBrende:
@molefrog sure thing, I can work on that.

HansBrende:
@molefrog PR for `useSyncExternalStore` submitted! https://github.com/molefrog/wouter/pull/267

molefrog:
@HansBrende FYI I've decided to release it, the change has landed in 2.9.0.

PR #267, opened 2022-11-16
add useSyncExternalStore

### LINKED ISSUE #58: Support for search

### LINKED ISSUE #140: Expose navigate, not requiring to subscribe to location state

### LINKED ISSUE #232: Expose a `subscribe` function suitable for React's `useSyncExternalStore`
The only thing missing from the `use-location` package is a way to actually subscribe to location updates.
Let's say such a `subscribe` function was exported.

Then users could do the following:
```typescript
import {subscribe} from "wouter";
const getCurrentValueOfMySearchParam = () => ... extract some stuff from window.location

function MyComponent() {
    const currentSearchValue = React.useSyncExternalStore(subscribe, getCurrentValueOfMySearchParam);
}
```

The nice thing about this potential ability is that you do *NOT* re-render if something in the URL changes that you don't care about!!!

Background: 
`useSyncExternalStore` is React's replacement for its previous attempt at a hook for listening to external subscriptions, called `useMutableSource`, which in turn was a replacement for its previous package called `useSubscription` which was a replacement for the gist created by Brian Vaughn linked to in [this comment](https://github.com/molefrog/wouter/blob/cba34af73d01389be18d0ed275611336946c18fe/use-location.js#L38).

In fact, all of the code surrounding that comment can likely be replaced by a single call to `useSyncExternalStore`. 

`useSyncExternalStore` is part of React 18; however there is a [shim package](https://www.npmjs.com/package/use-sync-external-store/v/rc) available for previous versions (so I don't think that would break your 0-dependency achievement since it is technically part of React.)

HansBrende:
Fixes #232 and fixes #58. 
Also fixes #140.

Also, simplifies existing code with `useEvent` and fixes a couple bugs along the way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,7 @@ import {
   useState,
   forwardRef,
   useIsomorphicLayoutEffect,
+  useEvent
 } from "./react-deps.js";
 
 /*
@@ -38,23 +39,15 @@ export const useRouter = () => useContext(RouterCtx);
  * Part 1, Hooks API: useRoute and useLocation
  */
 
-export const useLocation = () => {
-  const router = useRouter();
-  return router.hook(router);
-};
+const _useLocation = (router) => router.hook(router);
+
+export const useLocation = () => _useLocation(useRouter());
 
 export const useRoute = (pattern) => {
-  const [path] = useLocation();
-  return useRouter().matcher(pattern, path);
-};
-
-// internal hook used by Link and Redirect in order to perform navigation
-const useNavigate = (options) => {
-  const navRef = useRef();
-  const [, navigate] = useLocation();
-
-  navRef.current = () => navigate(options.to || options.href, options);
-  return navRef;
+  const router = useRouter();
+  // avoid inadvertently calling useRouter twice:
+  const [path] = _useLocation(router);
+  return router.matcher(pattern, path);
 };
 
 /*
@@ -103,12 +96,12 @@ export const Route = ({ path, match, component, children }) => {
 };
 
 export const Link = forwardRef((props, ref) => {
-  const navRef = useNavigate(props);
-  const { base } = useRouter();
+  const router = useRouter();
+  const [, navigate] = _useLocation(router);
 
-  let { to, href = to, children, onClick } = props;
+  const { to, href = to, children, onClick } = props;
 
-  const handleClick = useCallback(
+  const handleClick = useEvent(
     (event) => {
       // ignores the navigation when clicked using right mouse button or
       // by holding a special modifier key: ctrl, command, win, alt, shift
@@ -124,18 +117,14 @@ export const Link = forwardRef((props, ref) => {
       onClick && onClick(event);
       if (!event.defaultPrevented) {
         event.preventDefault();
-        navRef.current();
+        navigate(to || href, props);
       }
-    },
-    // navRef is a ref so it never changes
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [onClick]
-  );
+    });
 
   // wraps children in `a` if needed
   const extraProps = {
     // handle nested routers and absolute paths
-    href: href[0] === "~" ? href.slice(1) : base + href,
+    href: href[0] === "~" ? href.slice(1) : router.base + href,
     onClick: handleClick,
     to: null,
     ref,
@@ -158,8 +147,8 @@ const flattenChildren = (children) => {
 };
 
 export const Switch = ({ children, location }) => {
-  const { matcher } = useRouter();
-  const [originalLocation] = useLocation();
+  const router = useRouter(), matcher = router.matcher;
+  const [originalLocation] = _useLocation(router);
 
   for (const element of flattenChildren(children)) {
     let match = 0;
@@ -181,11 +170,13 @@ export const Switch = ({ children, location }) => {
 };
 
 export const Redirect = (props) => {
-  const navRef = useNavigate(props);
+  const {to, href = to} = props;
+  const [, navigate] = useLocation();
+  const redirect = useEvent(() => navigate(to || href, props));
 
-  // empty array means running the effect once, navRef is a ref so it never changes
+  // redirect is guaranteed to be stable since it is returned from useEvent
   useIsomorphicLayoutEffect(() => {
-    navRef.current();
+    redirect();
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
   return null;
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ import {
   forwardRef,
   useIsomorphicLayoutEffect,
   useEvent,
+  useInsertionEffect,
 } from "./react-deps.js";
 
 /*
@@ -68,7 +69,7 @@ export const Router = ({ hook, matcher, base = "", parent, children }) => {
   // we use `useState` here, but it only catches the first render and never changes.
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
   const [value] = useState(() => updateRouter({})); // create the object once...
-  useIsomorphicLayoutEffect(() => {
+  useInsertionEffect(() => {
     updateRouter(value);
   }); // ...then update it on each render
 
--- a/index.js
+++ b/index.js
@@ -68,16 +68,18 @@ export const Router = ({ hook, matcher, base = "", parent, children }) => {
 
   // we use `useState` here, but it only catches the first render and never changes.
   // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
-  const [value] = useState(() => updateRouter({
-    // We must store base as a property accessor because effects
-    // somewhat counter-intuitively run in child components *first*!
-    // This means that by the time a parent's base is updated in the
-    // parent effect, the child effect has already run, and saw
-    // the parent's *previous* base during its own execution.
-    get base() {
-      return (value.parent || defaultRouter).base + value.ownBase;
-    }
-  })); // create the object once...
+  const [value] = useState(() =>
+    updateRouter({
+      // We must store base as a property accessor because effects
+      // somewhat counter-intuitively run in child components *first*!
+      // This means that by the time a parent's base is updated in the
+      // parent effect, the child effect has already run, and saw
+      // the parent's *previous* base during its own execution.
+      get base() {
+        return (value.parent || defaultRouter).base + value.ownBase;
+      },
+    })
+  ); // create the object once...
   useInsertionEffect(() => {
     updateRouter(value);
   }); // ...then update it on each render
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ const defaultRouter = {
 
 const RouterCtx = createContext(defaultRouter);
 
-// gets the closes parent router from the context
+// gets the closest parent router from the context
 export const useRouter = () => useContext(RouterCtx);
 
 /*
--- a/index.js
+++ b/index.js
@@ -58,7 +58,7 @@ export const Router = ({ hook, matcher, base = "", parent, children }) => {
   const updateRouter = (router, proto = parent || defaultRouter) => {
     router.hook = hook || proto.hook;
     router.matcher = matcher || proto.matcher;
-    router.base = proto.base + base;
+    router.ownBase = base;
 
     // store reference to parent router
     router.parent = parent;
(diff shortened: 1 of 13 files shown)


HansBrende:
@molefrog any thoughts or feedback on this?

preact/react-deps.js
```diff
@@ -14,6 +14,8 @@ export {
   useCallback,
 } from "preact/hooks";
 
+export { useSyncExternalStore } from "preact/compat";
```
molefrog, line 17: A bit concerned here that we're introducing a dependency here on preact/compact, I left a [comment regarding that](https://github.com/molefrog/wouter/pull/259#pullrequestreview-1161000871) when we were working on `forwardRef`. The implementation itself is simple and it'll be tree-shaken, but we're locking our preact version to `^10.11.3` which is only 4 days old. Perhaphs, it makes sense to bring the shim inside of the `preact/react-deps.js` file. We can drop it along with shim for React in future releases of course. What do you think?

use-location.js
```diff
@@ -9,55 +13,46 @@ const eventPushState = "pushState";
...
+    }
+  };
+};
 
-    // it's possible that an update has occurred between render and the effect handler,
-    // so we run additional check on mount to catch these updates. Based on:
-    // https://gist.github.com/bvaughn/e25397f70e8c65b0ae0d7c90b731b189
-    checkForUpdates();
+const currentSearch = () => location.search;
+export const useSearch = () =>
```
molefrog, line 28: 👍

use-location.js
```diff
@@ -9,55 +13,46 @@ const eventPushState = "pushState";
...
+// that allows to perform a navigation.
+//
+// the function reference should stay the same between re-renders, so that
+// it can be passed down as an element prop without any performance concerns.
+export const useNavigate = (base = "") => {
+  const [nav] = useState([base, (to, opts) => navigate(to, opts, nav[0])]);
+  useIsomorphicLayoutEffect(() => {
+    nav[0] = base;
+  });
+  return nav[1];
```
molefrog, line 51: Could we rewrite it to just:

```js
export const useNavigate = (base = "") => 
  useCallback((to, opts) => navigate(to, opts, base), [base])
```

There must be some reason to use your version, but I just can figure out what it is 😄

preact/react-deps.js
HansBrende: @molefrog sure, that's fine with me. I don't know enough about Preact to know if the React shim is compatible with Preact since it has a peer dependency on React, not Preact. (I don't use Preact at all so you'll have to tell me.) Is that the case?

If so, great.

If not, and if we wanted to use an older version of Preact, we could go back as far as Preact 10.9.0 (the first release to include it), however, that version does have a buggy implementation of `useSyncExternalStore`. The reason I put it at 10.11.3 was because I noticed that release included their latest bugfix for `useSyncExternalStore`. Why they didn't just copy React's shim implementation in the first place is beyond me.

use-location.js
HansBrende: @molefrog yes, in fact there 2 reasons: 

1. My version won't unmemoize every time `base` changes, causing unnecessary re-rendering. The function returned from this hook is guaranteed to remain constant for the entire lifespan of this hook. (And in fact, I don't believe that even `useCallback` with an empty dependency array guarantees this since I believe it is implemented in terms of `useMemo`).

2. Since we are passing base in from a router, which is effectively a _ref_ (value that is mutated without triggering a re-render), that means that we are depending on the value of a ref in the callback dependency array, which you are not supposed to do, or at least it is nowhere documented that you can do this safely so it makes me very uneasy. See, for example: https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom

HansBrende: Ooh, but in talking about this, I just realized that I do need to make one small change here 😅 😅 😅

molefrog: Regarding 2), maybe I'm missing something afaik React compares dependencies using `===` and since `base` is supposed to be always a string the callback will be up to date.

preact/react-deps.js
molefrog: Gotcha. I would copy that latest version of the hook to the source code then. It's actually identical to what we had before, so I don't think we could introduce any regression.

use-location.js
HansBrende: @molefrog true perhaps, but the navigate function itself won't be up to date... since changing base does not trigger a re-render, it's possible that this callback won't ever be updated, leading to a situation where clicking a button, for example, sends you to the wrong path since the navigate function refers to `base` _as it was when the callback was memoized_, but not as it is now. That was actually the thing I fixed after saying "I do need to make one small change here".

HansBrende: And to be clear: when I say "changing base does not trigger a re-render", I'm referring to the re-render of child components within that component tree that are potentially using this navigate hook. Anything that is Memo'd won't re-render as a result of the re-render of the ancestor due to base changing in the first place. Hopefully that makes sense.

preact/react-deps.js
HansBrende: @molefrog done!

molefrog:
We probably need to modify our cjs config to tell rollup that `use-sync-external-store-shim` is an external dep. See `package.json`

HansBrende:
@molefrog I must admit, I am not exactly sure how to do that... do you have any guidance on how to proceed in that regard?

use-location.js
HansBrende: Now, one option that I considered is similar to what you suggest:

```typescript
export const useNavigate = (opts) => 
  useCallback((to, o) => navigate(to, o, opts.base || ""), [opts]);
```

That would work just fine. However it would fail for cases where the user is constructing `opts` dynamically, e.g.:
`useNavigate({base})` (fails) ❌   
`useNavigate(useRouter())` (works) ✅ 

Which is the reverse of the current logic in this library:
`useNavigate({base})` (works--provided `base` is a prop, not a ref) ✅ 
`useNavigate(useRouter())` (fails) ❌   

And thus might present backwards-incompatibility issues.

So instead of that, I opted for my current strategy which is the only one that works for _both_ usage patterns.

HansBrende: @molefrog any thoughts on the above?

use-location.js
```diff
@@ -17,55 +21,51 @@ const eventPushState = "pushState";
...
+const currentSearch = () => location.search;
+export const useSearch = () =>
+  useSyncExternalStore(subscribeToLocationUpdates, currentSearch);
 
-    return () => events.forEach((e) => removeEventListener(e, checkForUpdates));
-  }, [base]);
+export const usePathname = (opts = {}) =>
+  useSyncExternalStore(subscribeToLocationUpdates, () =>
+    relativePath(opts.base || "")
+  );
```
molefrog, line 42: Can we keep the current way of using default options?

```suggestion
export const usePathname = ({ base = "" } = {}) =>
  useSyncExternalStore(subscribeToLocationUpdates, () => relativePath(base));
```

I know that your version handles cases like `base = false` or `null` but I don't think we should allow users to pass that as an option.

use-location.js
molefrog: Fair point, think we can resolve this! 👍

molefrog (changes requested):
It seem like the size-limit check is failing, probably because we now have a polyfill. We have two options here:
1. Increase the size limit in `package.json`
2. Or, exclude `react-deps.js` from the size-limit bundle.

I personally prefer the second option because technically polyfill isn't part of the library.

use-location.js
HansBrende: @molefrog I was **not** actually doing that to support false or null, but rather to support `usePathname(router)` (since `base` may change inside that object, so it is important to look up the current value of base inside that object rather than copying the initial value supplied.

However... the option is actually a bit unsound in the first place, since `useSyncExternalStore` won't respond to base changes at all, thus if the base changes but the path does not, `usePathname(router)` won't work as expected. So I am going to modify this behavior to remove that option altogether from `usePathname` to limit the surface area of that behavior.

HansBrende:
@molefrog FYI: just noticed that this `useNavigate` function can be dramatically simplified by writing it in terms of the forthcoming React hook [useEvent](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md).

I will rewrite it that way now, to ease migration once `useEvent` comes out.

HansBrende:
@molefrog I tried excluding `react-deps.js` from the size-limit bundle as you suggested but it doesn't seem to be working. Any ideas?

index.js
```diff
@@ -38,23 +37,15 @@ export const useRouter = () => useContext(RouterCtx);
  * Part 1, Hooks API: useRoute and useLocation
  */
 
-export const useLocation = () => {
-  const router = useRouter();
-  return router.hook(router);
-};
+const _useLocation = (router) => router.hook(router);
```
molefrog, line 40: Thanks for updating this file as well. Do you think we could give this function a different name to avoid confusion? 
Maybe `useRouterLocation` (not ideal though) or `useLocationWithinRouter`?
As a well as some short comment to explain that it's an internal version of `useLocation` that works with any router passed to it.

molefrog (approved):
Looks good! Are we ready to merge this?

HansBrende:
@molefrog not quite yet! Need to add some typing updates.

HansBrende:
@molefrog alright, I believe this is ready to go. Just one question on [line 28](https://github.com/molefrog/wouter/blob/main/types/ts4.1/index.d.ts#L28):

```typescript
export * from "../use-location";
```

is that right?? I don't think we are re-exporting everything in `use-location` from index.js, are we? How should I proceed in that regard?

types/use-location.d.ts
```diff
@@ -26,16 +26,25 @@ export type HookNavigationOptions<H extends BaseLocationHook> = HookReturnValue<
...
+type Primitive = string | number | bigint | boolean | null | undefined | symbol;
+export const useLocationProperty: <S extends Primitive>(fn: () => S) => S;
+
+export const useSearch: () => string;
+
+export const usePathname: () => Path;
+
+export type Navigate = (to: Path, options?: { replace?: boolean }) => void;
+
+export const useNavigate: (options?: { base?: Path }) => Navigate;
```
HansBrende, line 38: @molefrog only thing I'm not sure on... am I supposed to use `export const` or `declare const` here (like was done below for `useLocation`)? I don't know what the difference is. (And I can't even find any documentation on what `declare const` does 😅 ).

molefrog: Good question. Based on what I was able to find, there is no actual difference between these two in a declaration file. There would be though if it was written in a source file. I think we stick to `export const` which is more correct in my opinion.

HansBrende: @molefrog cool, in that case, I think this is ready to be merged in if you don't have any further feedback! I'm happy with it.

HansBrende:
@molefrog alright, if you are good with it, I think this is ready to be merged in.

In addition to fixing the original issue I raised, #232, (via `useLocationProperty` to create an arbitrary subscription to any part of the url), it also fixes #58 as follows:

```typescript
import { useSearch, useLocationProperty, navigate } from 'wouter/use-location';

// get all search params:
const search = useSearch();

// set all search params:
navigate("?name=john");

// get individual search param (will not trigger a re-render if other params change! 🎉 )
const nameValue = useLocationProperty(() => new URLSearchParams(window.location.search).get("name"));

// set individual search param:
const params = new URLSearchParams(window.location.search);
params.set("name", "john");
navigate("?" + params.toString());
```

It also fixes #140 by exposing the global `navigate` function I've demonstrated above.

Feel free to merge at your convenience.

HansBrende:
@molefrog while waiting, I fixed 2 issues I noticed with the generated cjs build.

Let me know if you have any thoughts or comments. Otherwise, would love to see this merged in so I can finally put this PR behind me! 😅

molefrog:
Hey @HansBrende, happy new year to you!
I've just published `v2.10.0-alpha.1`. There's been quite some changes in the code so let's give it a try for a few weeks in real projects, and if everything is ok we can plan the next minor release.

PR #268, opened 2022-11-16
Support base path with the static location hook

### LINKED ISSUE #266: Perhaps the static location hook should support base path by default

roj1512:
- Closes #266.


static-location.js
```diff
@@ -1,16 +1,20 @@
+import { currentPathname } from "./utils.js";
+
 // Generates static `useLocation` hook. The hook always
 // responds with initial path provided.
 // You can use this for server-side rendering.
-export default (path = "/", { record = false } = {}) => {
```
molefrog, line 4: Unfortunately, this won't work since `base` is passed to the hook itself, not the factory. We have recently published support for parent routers in #265, and it assumes that base is always sent to the hook.

static-location.js
```diff
@@ -1,16 +1,20 @@
...
       if (replace) {
         hook.history.pop();
       }
-      hook.history.push(to);
+      // handle nested routers and absolute paths
+      hook.history.push(to[0] === "~" ? to.slice(1) : base + to);
     }
   };
+  path = currentPathname(path);
   hook = () => [path, navigate];
```
molefrog, line 18: I'm not sure yet to fix the history, but this would probably work as:

```suggestion
  hook = ({ base }) => [currentPathname(base, path), navigate];
```

static-location.js
roj1512: What won’t work?

molefrog: Or maybe we have different views on how this is going to work. In my understanding, the purpose of base path in static location is to produce the same output when you swap your browser hook with static one for Server-Side Rendering. 

Consider this app:
```js
import useBrowserLocation from "wouter/use-location";

<Router hook={useBrowserLocation} base="/app">
  <Route path="/users">
    <Link to="/home" /> {/* real path is "/app/home" *}
   </Route>
</Router>
```

And then let's say you'd like to swap your location hook with a static one to be able to render a specific page from the server:

```js
import staticLocation from "wouter/static-location";

// Important: `staticLocation` is not a hook just yet, it is a factory to produce the desired static hook
const useStaticLocation = staticLocation("/users")

<Router hook={useStaticLocation} base="/app">
  <Route path="/users">
    <Link to="/home" /> {/* in the markup the href is "/app/users" *}
   </Route>
</Router>
```

molefrog: In wouter the consideration about location hooks is that they are functions that may or may not accept an object with options and must return a pair of location and location setter.

`type LocationHook = (opts?: {}) => [string, (newLocation: string) => void]`

Wouter always sends `base` when it calls provided `useLocation``

roj1512: How is it now?

molefrog:
I think we should be consistent on what we write in the history. In my opinion, the history should always keep track of absolute paths (similar to the address bar in the browser), while the hook should return the relative path.

I've put together this specification to illustrate what I mean:

```
// <Router hook={staticLocation("/app")} base="/app" />
//
// (events)                (location)     (history)
// first render            "/"            -> ["/app"]
// navigate("/users")      "/users"       -> ["/app", "/app/users"]
// navigate("~/top")       "~/top"        -> ["/app", "/app/users", "/top"]

// <Router hook={staticLocation("/")} base="/app" />
//
// (events)                (location)     (history)
// first render            "~/"           -> ["/"]
// navigate("/users")      "/users"       -> ["/", "/app/users"]
// navigate("~/top")       "~/top"        -> ["/", "/app/users", "/top"]
```

Does this make sense? If so, I think we should write some basic unit tests to cover this new functionality and only then work on the implementation itself.

P.S. We do want to support absolute paths!

roj1512:
@molefrog How does it look now?

molefrog:
> @molefrog How does it look now?

The code breaks the tests, please make sure all tests are passed before committing. Let's also add a simple test case to test the added base path feature.

roj1512:
Sure. Also note that I am not a Node.js user so I cannot see what tests pass and which don’t before committing.

molefrog:
> Sure. Also note that I am not a Node.js user so I cannot see what tests pass and which don’t before committing.

Interesting. Are you using Deno? Do you have npm installed? If so, you can simply run `npm install` and then `npm test`

roj1512:
I just don’t feel like installing NPM. Sorry about that.

roj1512:
But don’t worry, we won’t leave this PR behind.

roj1512:
@molefrog I added a test and made the tests pass. How is everything now?

molefrog (approved):
🎉

PR #249, opened 2022-06-24
Add missing module specifiers to typesVersions

fgnass:
Currently, TypeScript can't find the type declarations for the module specifiers listed in the exports map. This PR fixes this.

PR #272, opened 2022-11-28
Fixes for wouter-preact

fgnass:
This PR is a follow-up to #249 to fix `wouter-preact` in native ESM environments. It does ... 

1. Add module specifiers from the exports map to typesVersions in `wouter-preact`, too.
2. Set the version to `"*"` for these exports, so it works in TS < 4.1, too.
3. Add missing `.js` extension, so the import can be resolved in native ESM envs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,10 +41,7 @@
   "module": "index.js",
   "types": "types/ts3.9.4/index.d.ts",
   "typesVersions": {
-    ">=4.1": {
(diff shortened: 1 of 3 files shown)


molefrog:
That's awesome! Have you already tested these changes in a ESM env?

molefrog (approved):
LGTM

PR #274, opened 2022-12-09
Fix missing parent Router declarations

### LINKED ISSUE #270: Missing type declaration for the new Router prop `parent`
https://github.com/molefrog/wouter/blob/main/types/ts4.1/index.d.ts#L125

I'm getting a ts error saying that `parent` doesn't exist on the Router type, which makes sense because it isn't declared in RouterProps. Unfortunately I am too new to typescript to be confident in contributing a fix.

molefrog:
Fixes #270

### DIFF
--- a/types/ts3.9.4/index.d.ts
+++ b/types/ts3.9.4/index.d.ts
@@ -93,8 +93,7 @@ export const Switch: FunctionComponent<SwitchProps>;
  * Components: <Router />
  */
 
-export type RouterProps = Partial<RouterOptions> & {
-  parent?: RouterObject;
+export type RouterProps = RouterOptions & {
   children: ReactNode;
 };
 
--- a/types/ts3.9.4/index.d.ts
+++ b/types/ts3.9.4/index.d.ts
@@ -29,7 +29,13 @@ export * from "../router";
 // React <18 only: fixes incorrect `ReactNode` declaration that had `{}` in the union.
 // This issue has been fixed in React 18 type declaration.
 // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210
-type ReactNode = ReactChild | Iterable<ReactNode> | ReactPortal | boolean | null | undefined;
+type ReactNode =
+  | ReactChild
+  | Iterable<ReactNode>
+  | ReactPortal
+  | boolean
+  | null
+  | undefined;
 
 /*
  * Components: <Route />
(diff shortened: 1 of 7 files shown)


types/ts3.9.4/index.d.ts
```diff
@@ -91,22 +93,18 @@ export const Switch: FunctionComponent<SwitchProps>;
...
-  base: Path;
-  matcher: MatcherFn;
-}
-export const Router: FunctionComponent<
-  Partial<RouterProps> & {
-    children: ReactNode;
-  }
->;
+export type RouterProps = Partial<RouterOptions> & {
+  parent?: RouterObject;
```
HansBrende, line 97: Shouldn't we move `parent` into `RouterOptions`? This would allow us to do `useRouter().parent`.

molefrog: 👍 Fixed in [1843f4c](https://github.com/molefrog/wouter/pull/274/commits/1843f4cd33f46c65fc9adfe4de96d626532e77e1)

types/ts4.1/index.d.ts
```diff
@@ -122,22 +125,18 @@ export const Switch: FunctionComponent<SwitchProps>;
...
-}
-export const Router: FunctionComponent<
-  Partial<RouterProps> & {
-    children: ReactNode;
-  }
->;
+export type RouterProps = Partial<RouterOptions> & {
+  parent?: RouterObject;
+  children: ReactNode;
+};
```
HansBrende, line 131: @molefrog `RouterOptions` is already `Partial` and includes `parent` now. So you can just do the same thing as for TS 3.9:

```typescript
export type RouterProps = RouterOptions & {
  children: ReactNode;
};
```

You could even move the definition of `RouterProps` into the shared file since it should be exactly the same as in 3.9 if I'm not mistaken.

molefrog: You are absolutely right. I must have overlooked that when I was refactoring the options. As for the shared file, yes indeed that would be nice, supporting duplicating types becomes a burden. I have some plans on refactoring that and providing a more elegant support for 3.9.

HansBrende: @molefrog `parent` can also be removed since that is already included in `RouterOptions`. Otherwise, looks good to me!

HansBrende (approved):
LGTM!

Issue #275, opened 2023-01-05
Can't resolve 'use-sync-external-store/shim' in `~/.../node_modules/wouter'. Did you mean 'index.js'?


jrnkng:
When I import `Route` from 'wouter' I get the following error.

```
Can't resolve 'use-sync-external-store/shim' in `~/.../node_modules/wouter'
Did you mean 'index.js'?
BREAKING CHANGE: The request 'use-sync-external-store/shim' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 ```

Created this project two days ago with create-react-app. Nothing special added really but somehow it doesn't work.

This is my `package.json`:
```
{
  "name": "...",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "@emotion/react": "^11.10.5",
    "@emotion/styled": "^11.10.5",
    "@mui/icons-material": "^5.11.0",
    "@mui/material": "^5.11.3",
    "@mui/styled-engine-sc": "^5.11.0",
    "@testing-library/jest-dom": "^5.16.5",
    "@testing-library/react": "^13.4.0",
    "@testing-library/user-event": "^13.5.0",
    "@types/jest": "^27.5.2",
    "@types/node": "^16.18.11",
    "@types/react": "^18.0.26",
    "@types/react-dom": "^18.0.10",
    "react": "^18.2.0",
    "react-bootstrap-icons": "^1.10.2",
    "react-dom": "^18.2.0",
    "react-dropzone": "^14.2.3",
    "react-scripts": "5.0.1",
    "styled-components": "^5.3.6",
    "typescript": "^4.9.4",
    "web-vitals": "^2.1.4",
    "wouter": "^2.10.0-alpha.1"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject"
  },
  "eslintConfig": {
    "extends": [
      "react-app",
      "react-app/jest"
    ]
  },
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  }
}
```

EDIT:

[...]

PR #277, opened 2023-01-12
fix: update module path to fully qualified name

Ziinc:
improves fix for #275

### DIFF
--- a/react-deps.js
+++ b/react-deps.js
@@ -21,7 +21,7 @@ export {
 // and https://github.com/TanStack/query/pull/3601
 // ~ Show this aging PR some love to remove the need for this hack:
 //   https://github.com/facebook/react/pull/25231 ~
-export { useSyncExternalStore } from "./use-sync-external-store";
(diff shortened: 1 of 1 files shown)


HansBrende:
@molefrog @Ziinc this won't work on native... the reason we need to use unqualified extensions is so that the bundler can decide which extension to pick up, `.native.js`, or `.js`.

HansBrende:
Ok, I looked at the generated build files from react query... it looks like their generated ESM files do a:

```javascript
import {useSyncExternalStore} from './useSyncExternalStore.js'
```

so we should be good there (I think). However, in their cjs files, they are `requiring` the unqualified names, e.g.
```javascript
require('./useSyncExternalStore')
```
so native may not work for us in those cases. They accomplish this via the following:
https://github.com/TanStack/query/blob/86161ca6ec7b29f71576b555206aca04831499e6/rollup.config.ts#L374

molefrog:
I've raised a PR to have an external `rollup.config.mjs` file to allow plugins.

PR #280, opened 2023-01-20
Create external Rollup config

molefrog:
The build command in package.js has become quite messy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wouter",
-  "version": "2.10.0-alpha.1",
+  "version": "2.10.0-alpha.4",
   "description": "A minimalistic routing for React and Preact. Nothing extra, just HOOKS.",
   "keywords": [
     "react",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -68,9 +68,9 @@
     "lint-types": "npm run lint-types-react && npm run lint-types-preact",
     "lint-types-react": "dtslint types/ts4.1 --localTs ./node_modules/typescript/lib && dtslint types/ts3.9.4 --localTs ./node_modules/typescript-3.9.4/lib",
     "lint-types-preact": "npm run preact && npm run types-preact && dtslint preact/types/ts4.1 --localTs ./node_modules/typescript/lib && dtslint preact/types/ts3.9.4 --localTs ./node_modules/typescript-3.9.4/lib",
-    "preact": "copyfiles -f {index,matcher,use-location,static-location}.js preact/",
+    "preact": "copyfiles -f {index,matcher,use-location,static-location,use-sync-external-store,use-sync-external-store.native}.js preact/",
```
HansBrende, line 71: @molefrog I don't think Preact uses `use-sync-external-store`! The Preact implementation is baked into `preact/react-deps`.

molefrog: You're right! I must have overlooked that. Fixed in https://github.com/molefrog/wouter/pull/280/commits/275b42baebf5a8ad8a3f9816bb42df9ef9dcf1f5

molefrog: We could have used `PREACT=true` env var, but we needed to output `package.json` in the build script, and I've decided not to overcomplicate things.

PR #285, opened 2023-02-02
Support new moduleResolution of TypeScript >= 4.7.

toerndev:
TypeScript 4.7 adds new values for "moduleResolution": "node16" and "nodenext", which enable ESM features.

Since `exports` now takes precedence in package.json and wouter's type files are not located next to their corresponding .js-files, it is necessary to declare `types` for each entry under `exports`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,18 +21,22 @@
   "type": "module",
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


molefrog (approved):
Thanks!

Issue #286, opened 2023-02-07
`useSyncExternalStore` issue with SSR


mashaal:
When using v2.10.0 with SSR and React 18, we are getting a React invariant 407 error.

`Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.`

https://reactjs.org/docs/error-decoder.html/?invariant=407

HcroakerDev:
Getting the exact same issue myself. Anyone know any fixes??

mashaal:
> Getting the exact same issue myself. Anyone know any fixes??

We've temporarily downgraded to `2.9.0` before the `useSyncExternalStore` was added

cbbfcd:
<img width="497" alt="image" src="https://user-images.githubusercontent.com/20987775/217419216-37266823-2df9-413a-8def-97fb82613c8f.png">

maybe we should add an empty function for `getServerSnapshot `.

molefrog:
> When using v2.10.0 with SSR and React 18, we are getting a React invariant 407 error.
> 
> `Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.`
> 
> https://reactjs.org/docs/error-decoder.html/?invariant=407

Thanks for reporting. Is this coming from the client or the server? In theory, you shouldn't get this on the server because `static-location` doesn't use `useSyncExternalStore`...

molefrog:
My guess is that, React notices an inconsistency between what's been rendered on the server and a the hydrated variant. Luckily, `useSyncExternalStore` has native support for SSR so I assume we can drop static location entirely now! We just have to tweak the API a little bit. I'm going to run a simple experiment and will get back to you with the answer.

WavyWalk:
Facing same issue. Error happens during hydration. (nice work btw)

molefrog:
Hi, I've been able to track down the issue. We switched to `useSyncExternalStore` API in the latest release but forgot to support the 3rd argument that defines the value returned from the hook when it is invoked on the server. This is actually a big improvement for us, since it'll allow us to stop using `static-location` on the server.

However, our current `wouter/use-location` isn't extendable and we'll have to spend some time tweaking the API to enable SSR support. In the meantime, here is a workaround that you can basically copy into your project: https://gist.github.com/molefrog/70ad57db1acf689b3f4eaf77b7065233

It's not ideal (there are some methods that I had to copy from the original file), but it's good enough until we release the fix.

Sorry for the inconvenience!

[...]

PR #305, opened 2023-05-01
SSR support in `syncExternalStore`

molefrog:
See #286 

TODO:
- ~~README~~
- ~~Publish a test version~~


molefrog:
Merging this as the API is more or less stable at this point.

PR #312, opened 2023-05-20
update packages sizes

felixcatto:
I think minified versions is better for size comparison. Because wouter becomes only 2x smaller after gzip, but react-router - 3x smaller.

And i think minified size matters even more than gzipped, because scripts loads once, then cached. But they parsed and 
evaluated on every load. Therefore 65 Kb slow down script execution much more than 4.4 Kb

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 <img src="assets/wouter.svg" align="right" width="250" alt="by Katya Simacheva" />
 
-- Zero dependency, only **1.36 KB** gzipped vs 11KB
+- Zero dependency, only **4.4 KB** minified vs 65.8KB
   [React Router](https://github.com/ReactTraining/react-router).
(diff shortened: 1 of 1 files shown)


molefrog:
That's fair enough. I would maybe put it this way though: "Only 1.5KB gzipped (4.5KB minified)".
For comparison, Preact advertises itself as "3KB alternative for React", while their minified version is over 10KB.

We could also solve this by having a more detailed table comparing RR and wouter in all possible forms: original, minified, minified + gzipped. 

image

felixcatto:
I just wanted to update info about sizes, because the difference between React-router now is bigger. Feel free to discard this pull request and write whatever you think is more appropriate :wink:

molefrog:
> I just wanted to update info about sizes, because the difference between React-router now is bigger. Feel free to discard this pull request and write whatever you think is more appropriate 😉

Gotcha! Thank you for suggesting this, you have a really good point here.

PR #314, opened 2023-05-22
Deduplicate code

jeetiss:
Tweak rollup config to deduplicate path helpers in code base.


rollup.config.mjs
```diff
@@ -18,7 +20,7 @@ const ESM_SOURCES = [
 
 const OUTPUT_DIR = "cjs";
 
-export default [
+export default defineConfig([
```
molefrog, line 23: what's the advantage of using `defineConfig` here?

static-location.js
```diff
@@ -1,13 +1,4 @@
...
-// that will also pull in react, use-sync-external-store, and then
-// monkeypatch `history` *again* in the generated build files!
-const relativePath = (base = "", path = location.pathname) =>
-  !path.toLowerCase().indexOf(base.toLowerCase())
-    ? path.slice(base.length) || "/"
-    : "~" + path;
-
-const absolutePath = (to, base = "") =>
-  to[0] === "~" ? to.slice(1) : base + to;
+import { absolutePath, relativePath } from "./path-functions.js";
```
molefrog, line 1: Would be nice to remove the word `function` from the module name. Maybe, `paths.js` or `base-path.js`?

rollup.config.mjs
jeetiss: `defineConfig` provides type definitions:

Снимок экрана 2023-05-23 в 09 44 25

PR #318, opened 2023-05-26
Migrate packages to monorepo

jeetiss:
- migrate `wouter` and `wouter-preact` to monorepo
- export only `esm`, drop `cjs` exports
- test with `jest` (but coverage is  broken)


package.json
```diff
@@ -1,90 +1,23 @@
...
-    "preact": "copyfiles -f {index,matcher,use-location,static-location,paths}.js preact/",
-    "types-preact": "copyfiles -f types/{matcher,use-location,static-location,router}.d.ts preact/types",
-    "bundle": "rollup -c && echo '{\"type\": \"commonjs\"}' > ${DIR}cjs/package.json",
-    "prepublishOnly": "npm run clean && npm run preact && npm run types-preact && npm run bundle && DIR=./preact/ npm run bundle"
+    "lint-types-react": "dtslint packages/wouter/types/ts4.1 --localTs ./node_modules/typescript/lib && dtslint packages/wouter/types/ts3.9.4 --localTs ./node_modules/typescript-3.9.4/lib",
+    "lint-types-preact": "npm run types-preact && dtslint packages/wouter-preact/types/ts4.1 --localTs ./node_modules/typescript/lib && dtslint packages/wouter-preact/types/ts3.9.4 --localTs ./node_modules/typescript-3.9.4/lib",
+    "types-preact": "copyfiles -f packages/wouter/types/{matcher,use-location,static-location,router}.d.ts packages/wouter-preact/types",
+    "prepublishOnly": "npm run build && npm run types-preact"
```
molefrog, line 20: this command is not needed in this package, it won't be published anyway

package.json
```diff
@@ -103,18 +36,26 @@
   },
   "size-limit": [
     {
-      "path": "index.js",
-      "limit": "1500 B",
-      "ignore": [
-        "./react-deps.js"
-      ]
+      "path": "packages/wouter/dist/index.js",
+      "limit": "2000 B",
```
molefrog, line 40: Limits are increased, may I ask why?

molefrog (changes requested):
In the root, there is a file `paths.js` I guess we don't need it anymore.

molefrog (changes requested):
In the root, there is a file `paths.js` I guess we don't need it anymore.

packages/wouter-preact/rollup.config.js
```diff
@@ -0,0 +1,28 @@
...
+    input: [
+      "wouter",
+      "wouter/matcher",
+      "wouter/static-location",
+      "wouter/use-location",
+    ],
+    external: ["preact", "preact/hooks"],
+
+    output: {
+      dir: "dist",
```
molefrog, line 16: Can we rename the folder to `esm`? I've seen some libraries doing that, I think it adds just a little bit of clarity.

package.json
jeetiss: to make `size-limit` happy)

I included `react-deps.js` to bundle-size test and it not only increases the size but causes some deoptimizations i guess:

- https://github.com/molefrog/wouter/blob/e5544caa1a9059ff24da4fee4893021068541f34/packages/wouter/src/react-deps.js#L4-L9

this code can be deleted and included in final size tests

jeetiss: it should be 

```js
import { 
   useEffect, 
   useLayoutEffect, 
   useRef, 
   useInsertionEffect as useBuiltinInsertionEffect, 
 } from 'react'; 
```

to fix all three-shaky stuff

jeetiss: but it can break some old react versions that doesn't provide `useInsertionEffect`

molefrog (approved):

PR #320, opened 2023-05-31
base vitest setup

jeetiss:
- setup `vitest`
- move tests to `typescript`
- ~add a lot of `// @ts-expect-error` comments~


packages/wouter-preact/package.json
```diff
@@ -35,12 +35,11 @@
       "default": "./esm/static-location.js"
     }
   },
-  "module": "esm/index.js",
-  "types": "types/ts3.9.4/index.d.ts",
+  "types": "types/index.d.ts",
   "typesVersions": {
```
molefrog, line 39: Could you check what versions of TS support type declaration paths defined in `exports`? I've got a feeling that we don't need `typesVersions` at all now.

packages/wouter-preact/test/preact.test.tsx
```diff
@@ -33,15 +33,15 @@ describe("Preact support", () => {
 
         <main data-testid="routes">
           <Switch>
-            Welcome to the list of {100} greatest albums of all time!
+            <>Welcome to the list of {100} greatest albums of all time!</>
             <Route path="/albums/all">Rolling Stones Best 100 Albums</Route>
-            <Route path="/albums/:name">
+            <Route<{ name: string }> path="/albums/:name">
```
molefrog, line 38: Hmm, routes should infer the proper type of parameters from the path.

packages/wouter-preact/package.json
jeetiss: nope, `type: module` doesn't affect how typescript resolves types. `typesVersions` is used when `moduleResolution` is set to `node`, and the `type` field in `exports` is used when `moduleResolution` is `nodenext` or `node16` or `bundler`

molefrog: ok!

molefrog (approved):
Looks good! Could you pls check why route parameters are not inferred?

packages/wouter/test/use-location.test.tsx
```diff
@@ -31,16 +32,17 @@ describe("`value` first argument", () => {
     unmount();
   });
 
-  it("supports history.back() navigation", () => {
-    jest.useFakeTimers();
+  // Vitest doesn't support Jest's legacy timers.
+  it.skip("supports history.back() navigation", () => {
```
molefrog, line 36: I don’t even remember why I had to use fake timers here. Does the test pass when you remove these calls?

packages/wouter-preact/test/preact.test.tsx
jeetiss: I managed to fix this

packages/wouter/test/use-location.test.tsx
jeetiss: I replaced `FakeTimers` with `waitFor` util from `testing-library`

molefrog:
Nice, thank you!

Issue #316, opened 2023-05-24
be able to add state into the link like react router


numbing:
in react router we can add state into the Link that work like this:
`<Link to={{pathname: '/nextpath', state: { prevPath: location.pathname }}}>Example Link</Link>`

it would be great if had a feature that work like this in wouter

redwanul10:
Hi @numbing ! I have made a PR [#325](https://github.com/molefrog/wouter/pull/325)  let's wait for @molefrog to review :smiley:

molefrog:
Hey @redwanul10 thank you for your PR. We made a complete revamp of the library and state support is now available in v3.0.0-rc.1 (see my comment https://github.com/molefrog/wouter/issues/102#issuecomment-1822253686)

We also made a standalone hook for getting the state:

```js
import { useHistoryState } from "wouter/use-browser-location"

// this will only re-render when the actual state changes, not the URL
const state = useHistoryState()
```

It also uses `useSyncExternalStore` under the hood, so it solves the SSR issue you had.

PR #325, opened 2023-06-17
feat: add navigation state support to routes 

redwanul10:
in short: this PR is related to [#316](https://github.com/molefrog/wouter/issues/316)
### Details
Currently in wouter navigation state feature is not available, 
that's why we can't pass state to routes like **react-router-dom** . so I have added this feature. 
### Examples:
```js
import { Link , useLocation } from "wouter";

// pass state using Link component
<Link href="/foo" state={{ msg: "hello" }} className="active">Hello!</Link>

// get & set state using useLocation hooks
const [location, setLocation, state] = useLocation();

setLocation("/user", { state: { msg: "hello" } }); 
```

### DIFF
--- a/use-location.js
+++ b/use-location.js
@@ -40,8 +40,8 @@ export const usePathname = ({ ssrPath } = {}) =>
     ssrPath ? () => ssrPath : currentPathname
   );
 
-export const navigate = (to, { replace = false } = {}) =>
-  history[replace ? eventReplaceState : eventPushState](null, "", to);
+export const navigate = (to, { replace = false, state = null } = {}) =>
+  history[replace ? eventReplaceState : eventPushState](state, "", to);
 
 // the 2nd argument of the `useLocation` return value is a function
 // that allows to perform a navigation.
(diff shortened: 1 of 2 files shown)


redwanul10:
Hi @molefrog can you please let me know what should I need to do to work this in SSR ? I tested it works on browser but according to the test case it showing error on SSR !

use-location.js
```diff
@@ -26,6 +26,14 @@ const subscribeToLocationUpdates = (callback) => {
   };
 };
 
+const getHistoryState = () => {
+  try {
+    return history.state;
+  } catch (error) {
```
molefrog, line 32: Is wrapping this line in `try..catch` really needed? What are the cases when it could throw?

redwanul10: without **`try..catch`** it works totally fine on browser but throws error when you run the test cases `npm run test`

molefrog:
@redwanul10 Thank you for your help! I am closing this per https://github.com/molefrog/wouter/issues/316#issuecomment-1822264860.

Issue #245, opened 2022-05-06
Need a hook to get parameters - useParams


m4n1ac47:
You need to make a custom hook `useParams` to get the route parameters.

```js
import  { useParams }  from "wouter"

const Component = (props) => {
   const { id } = useParams()
  
   return (
     <div>{ id }</div>
  )
} 
```

[Your solution](https://github.com/molefrog/wouter/issues/161#issuecomment-759494019) that you provided in the comment is not suitable for complex cases, you have to constantly write the path. Sometimes it is complex and has a regular expression, and it is not a good idea to write it all the time.

```js
const Component = () => {
  const [, params] = useRoute("/user/:name");
  return <div>{params.name}</hi>
};
```

Your example is simple, now let's see how we can get the parameters from the given route.
`/en/product/gamepad-sony-dualshock-4-v2-cuh-zct2e-8`

```js
const Component = () => {
  const [, { id, lang }] = useRoute('(/)*:lang([a-z]{0,2})/product/([a-z0-9-_]*[-_]*)*:id')
  return <div>{id}</hi>
};
```

Also, this solution is not suitable when you need to access the parameters, when the component is universal and it does not know how the routes are described in the application.

cbbfcd:
This is a previous discussion item, I remember the point of disagreement is whether it needs to be provided by wouter or defined by the user.

m4n1ac47:
I think this should be available in the default library because it's a much needed feature. It will be much more convenient and better if this function is made by the Wouter authors, because third-party developments cannot guarantee the same level of quality and optimization. For some it will work well, for others there will be bugs.

And it's just nicer when many good ones are available out of the box.

cbbfcd:
maybe will be supported in wouter-next

yairEO:
There is a native browser API for this, why do you need this library to waste a single byte of code doing what is already
available out-of-the-box?

https://stackoverflow.com/a/9870540/104380

m4n1ac47:
You are talking about query params. I'm talking about route parameters. Try to take `/admin/panel/product-offer/4598/product/246` these parameters and understand what is what.
It should be in the library because it's convenient.

```js
const { offerId, productId } = useParams()
```

> There is a native browser API for this, why do you need this library to waste a single byte of code
> doing what is already available out-of-the-box?

yairEO:
what are "route params"? everything should always be in the window URL query so when the user refreshes, the state will persist

molefrog:
[...]

Issue #161, opened 2020-12-31
Make hooks for params (useParams)


CoffeJeanCode:
Make a custom hook to use the parameters, to make a better code and have more control of the parameters, and be able to have more flexibility when using them

```jsx
import  useParams  from "wouter/use-params"

const Component = (props) => {
   const parms = useParams( )
  // /user/:name
   return (
     <div>{ params.name  }</div>
  )
} 
```

omgovich:
Hi, @CoffeJeanCode,

`useRoute` doesn't solve the problem?

```jsx
const Component = () => {
  const [, params] = useRoute("/user/:name");
  return <div>{params.name}</hi>
};

```

CoffeJeanCode:
Thanks, you

m4n1ac47:
`useRoute` - doesn't solve the problem because you have to write a full regex in complex-dynamic routes. In a bunch of routes, this becomes a problem. Because it would not be desirable to write it regular expressions in all places.

```js
const Component = () => {
  const [, { id, lang }] = useRoute( '(/)*:lang([a-z]{0,2})/category/([a-z0-9-_]*[-_]*)*:id')
  return <div>{id}</hi>
};
```

PR #329, opened 2023-08-16
Add useParams hook

Vija02:
This PR adds support for a `useParams` hook, briefly discussed in #161 and #245.
The main benefit is to allow users to access this data in any child component without having to pass the data all the way down. This makes creating more generic components tidier.

### Example
```js
import { Route, useParams } from "wouter";

const User = () => {
  // This can be called anywhere within the Route component
  const params = useParams();

  console.log(params.id)
};

<Route path="/user/:id" component={User}> />
```

### How does this work?
We wrap the user's component with a context provider. The `useParams` function in a simple wrapper for `useContext`.

### More info
The full changes itself are just a few lines in `index.js` ~and a new file `use-params.js`.~

This can be further minimized if needed but I think that'll make it much less readable.

Checking the size using `yarn size` shows that `index.js` grew from `1.05 kB` -> `1.06kB`.

I've done my best to add tests & documentation but they're pretty basic so please feel free to make changes or add more of them. Also, I'm not sure how the types work in this project so I tried matching existing ones but please do check them. Thanks!


molefrog (changes requested):
Hi, thank you for working on adding this. The only thing I would change is I would move use-params.js inside of the main module. This method can't be used independently from the rest of the library, so there is no need to have an extra entry point.

Vija02:
Thanks for having a look @molefrog! I've added a commit to move everything in `use-params.js` into `index.js` as you suggested.

dominiceden:
Thanks @Vija02 - just wanted to say that this was the biggest thing I was missing vs React Router so looking forward to this! Thanks @molefrog for the great work on this lib too.

molefrog (approved):

molefrog:
@Vija02 I have published wouter@2.12.0-beta.1 
There are some other features that I am considering for the next release, but you can use this version in the meantime.

PR #336, opened 2023-09-13
V3: `regexparam`-based route matching

molefrog:
Removes default matcher and replaces it with [`regexparam`](https://github.com/lukeed/regexparam)

## Breaking changes

Matching
  - no plus sign modifiers `"/:user+"`
  - no named wildcards `"/:user*"`
  - no optional wildcards `"/*?"` (**yet!** see https://github.com/lukeed/regexparam/pull/25)
  - `"/users/:id-:name"` doesn't work. But, this is supported by `path-to-regexp`
  - when pattern ends with a trailing slash, it matches both `"/home/"` and `"/home"`, but I suppose it is a better DX
  - wildcards can only be present at the end of the string, e.g. `"/home/:rest*/all"` will not work
  - router now accepts `parser` prop that allows to customize pattern parsing. Now it is easier to integrate with `path-to-regexp`

Base paths
  - Routes can no longer refer to absolute paths (e.g. `~/home`). Why? Because when you are inside a nested context, the content won't be rendered anyway and it requires additional work to support it in `regexparam`

Modules
  - 'wouter/matcher' has been removed

## TODO
  - [ ] TypeScript types
  - [ ] Type specs for `useRoute`


molefrog:
Wrong branch, closing this.

PR #337, opened 2023-09-13
V3: `regexparam`-based route patterns

molefrog:
Removes default matcher and replaces it with [`regexparam`](https://github.com/lukeed/regexparam)

## Breaking changes

Matching
  - no plus sign modifiers `"/:user+"`
  - no named wildcards `"/:user*"`
  - no optional wildcards `"/*?"` (**yet!** see https://github.com/lukeed/regexparam/pull/25)
  - `"/users/:id-:name"` doesn't work. But, this is supported by `path-to-regexp`
  - when pattern ends with a trailing slash, it matches both `"/home/"` and `"/home"`, but I suppose it is a better DX
  - router no longer accepts `matcher` prop. Instead, there is a `parser` prop that allows to customize pattern parsing. Now it is easier to integrate with `path-to-regexp`

Base paths
  - Routes can no longer refer to absolute paths (e.g. `~/home`). Why? Because when you are inside a nested context, the content won't be rendered anyway and it requires additional work to support it in `regexparam`

Modules
  - `wouter/matcher` has been removed

## TODO
  - [x] TypeScript types
  - [x] Type specs for `useRoute`

## Further improvements
1. Memoize constructed patterns: don't parse the string when pattern doesn't change
2. `useRoute` should not work with falsey patterns, while `Route` should always render when pattern isn't provided


molefrog:
@jeetiss Hi, here are the things I need your help with:
- Review the PR. Did I miss anything?
- Make vitest ignore old type tests and drop `lint-types` command in favour of `test-types`

jeetiss (approved):
I dropped the cache and almost all tests work now, could you fix last one in `wouter-preact` package

https://github.com/molefrog/wouter/blob/3dd87a7fa658e488f4c482e7abdafd6b9656612d/packages/wouter-preact/test/preact.test.tsx#L41

(as I understand named wildcard params doesn't supported by parser)

i think it is better to disable tests for types now (or merge with red ci) and i going to try to fix them in separate PR

molefrog:
> i think it is better to disable tests for types now (or merge with red ci) and i going to try to fix them in separate PR

This test does work for me, how do I clean the cache?

molefrog:
I figured it out! The test has been fixed.

molefrog:
Apparently, there was no actual task for linting types on CI 😄

Issue #346, opened 2023-09-27
`use-browser-location`


molefrog:
For V3, we want to ship a standalone hook `wouter/use-browser-location` that will replace `wouter/use-location`. The primary motivation for this is to remove confusion around `import { useLocation } from "wouter"` and `import useLocation from "wouter/use-location"`.

This hopefully will make it more clear that the standalone hook can be used without wouter, while the hook imported from `"wouter"` will always rely on the current router's location. 

I suggest that we don't just rename this module, but also work on some things the current version lacks:
- [x] Ensure that we monkey-patch `pushState` only once, see #167 
- [x] Consider using `globalThis` when accessing `location` (this needs to be researched) #344
- [x] Allow to pass current `state` when performing navigation #102
- [x] **Breaking change:** I think we can safely drop this `export const events`, since it looks more like an implementation detail
- [x] **Breaking change:** Use named exports only, `import { useBrowserLocation } from "wouter/use-browser-location"`
- [x] Migrate type specs

molefrog:
Additional: does the `state` have to be returned from `useBrowserLocation`? There is a PR #325 that adds it as a third argument, however I am not sure it's the right approach, since other hooks (like hash location or memory) will not implement it.

We could however allow location hooks to return the optional object that they can populate with whatever implemetation-specific parameters they need.

molefrog:
@jeetiss Can we close this now?

PR #348, opened 2023-09-29
v3: use browser location

jeetiss:
first part of `use-browser-location` feature:

- Renamed `use-location`
- Named exports only
- Dropped `export const events` from module
- Migrated type specs

related to #346

### DIFF
--- a/packages/wouter/test/use-browser-location.test-d.ts
+++ b/packages/wouter/test/use-browser-location.test-d.ts
@@ -1,0 +1,15 @@
+import { it, assertType } from "vitest";
+import { useBrowserLocation } from "wouter/use-browser-location";
+
+it("should return string, function tuple", () => {
+  const [loc, navigate] = useBrowserLocation();
+
+  assertType<string>(loc);
+  assertType<Function>(navigate);
+});
+
+it("should support options", () => {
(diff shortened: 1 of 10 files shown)


packages/wouter/test/use-browser-location.test-d.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useBrowserLocation } from "wouter/use-browser-location";
+
+it("should return string, function tuple", () => {
+  const [loc, navigate] = useBrowserLocation();
+
+  assertType<string>(loc);
+  assertType<Function>(navigate);
+});
+
+it("should support options", () => {
```
molefrog, line 11: Can you specify which options it supports here?

packages/wouter/test/use-browser-location.test-d.ts
```diff
@@ -0,0 +1,15 @@
+import { it, assertType } from "vitest";
+import { useBrowserLocation } from "wouter/use-browser-location";
+
```
molefrog, line 3: Can you also add a spec to assert what parameters `navigate` accepts?

molefrog (approved):
LGTM! What are the next steps?

PR #349, opened 2023-10-01
v3: history state hook

### LINKED ISSUE #102: Discuss about the `state` pass into Link or navigate
Maybe we also need another way to pass parameters.

just like this:

```js
const [, navigator] = useLocation();
navigate(path: string, replace?: boolean, state?: object);

// and
<Link to='/app' state={state}/>
```

@omgovich @molefrog

jeetiss:
Api for `useHistoryState`, should solve all problems with `history.state`

----

usage example:

```js 

const Component = () => {
  const state = useHistoryState();

  // some logic with state
  // console.log(state) 
  // ^ { hello: 'world' }
}

// somewhere in the app
navigate('/path', { state: { hello: 'world' }})
```

---

TODOs:

- [x] types
- [x] link component
- [x] preact
- [x] check exports 

---

closes #102
related #346


packages/wouter-preact/types/use-location.d.ts
```diff
@@ -39,10 +39,11 @@ export const useSearch: () => string;
 
 export const usePathname: (options?: { ssrPath?: Path }) => Path;
 
-export const navigate: (
-  to: string | URL,
-  options?: { replace?: boolean }
-) => void;
+export const useHistoryState: <T = any>() => T;
```
molefrog, line 42: I think we can rename this file to `use-browser-location`, as well as move everything that goes before `export type Primitive` to `index.d.ts`

packages/wouter/src/use-browser-location.js
```diff
@@ -40,8 +40,11 @@ export const usePathname = ({ ssrPath } = {}) =>
     ssrPath ? () => ssrPath : currentPathname
   );
 
-export const navigate = (to, { replace = false } = {}) =>
-  history[replace ? eventReplaceState : eventPushState](null, "", to);
+const currentHistoryState = () => history.state;
+export const useHistoryState = () => useLocationProperty(currentHistoryState);
```
molefrog, line 44: We can safely assume that `state` in SSR is always `null`, so let's provide a second argument here.

packages/wouter/types/use-location.d.ts
```diff
@@ -53,7 +54,7 @@ export const navigate: (
 // navigate with `pushState` or `replaceState`.
 export type LocationHook = (options?: {
   base?: Path;
-}) => [Path, (to: Path, options?: { replace?: boolean }) => void];
+}) => [Path, (to: Path, options?: NavigateOptions) => void];
```
molefrog, line 57: The 2nd element in this tuple is `navigate`, so we can reuse its type here.

molefrog (changes requested):
Could you also add a one-liner spec to test that `Link` can accept `state` prop?

packages/wouter-preact/types/use-location.d.ts
jeetiss: I renamed the file.

but moving is not great because it is harder to maintain `preact` types

molefrog (approved):

PR #350, opened 2023-10-01
v3: Patch history once

### LINKED ISSUE #167: Discussion: leaky implementation for pushState/replaceState events
Hi!

Looking at the code for `useLocation`, I wonder if we can do something about the way `pushState` and `replaceState` events are currently handled. 

## Basic issue

Basically what I find problematic, is the fact that the monkey-patching is done eagerly. While I understand that this is needed for Wouter to correctly react to history changes triggered from outside Wouter, I wonder if we could somehow make this behavior opt-in (without having to rely on a custom `useLocation` hook)

## Real-world problematic use-case

The main use-case that bothers me is when using Wouter inside a Micro-frontend architecture such as a Unified Single Page Application.

If multiple microfrontends all bundle their own version of Wouter, the `replaceState` and `pushState` event will get monkey-patched several times, resulting in multiple events being fired at each call.

### Flagging 

We could maybe simply add a flag to `replaceState`/`pushState` (or even directly on `window`) to mark that they have been patched by Wouter. This way, subsequent Wouters won't patch them again.

The downside is that does not allow users to opt-out of the patching altogether, it really only focuses on the microfrontends use-case.

### Exposing a `pure` endpoint

This bit of code in `useLocation` is actually the only part of Wouter that makes it un-pure. Extracting it out of `useLocation`'s base code would allow us to provide two entry points. Either users import `"wouter"` or `"wouter/useLocation"` to get the default behavior, or they import `"wouter/pure"` or `"wouter/pure/useLocation"` to get the side-effect-free implementation.

The upside is that it allows true opt-out of the side-effect, so it covers all use-cases when such an opt-out is needed.

The downside is that it makes the export API a little more complex because of those new entry-points.
It might also make the code more complex, as right now Wouter's internal routing relies on those events too. With this solution we might want to add a call to `checkForUpdates` in the default `navigate` function.

### Making it pure by default

Finally, this last option is a bit more drastic. We could export a `setupEventListeners` function with this side-effect code, and have wouter/useLocation be pure by default. Users could then _opt-in_ for event listeners by calling this somewhere in their app.

The upside is that it allows true opt-out of the side-effect, so it covers all use-cases when such an opt-out is needed, while maintaining simple entry-points.

The downside is that _most users might want to opt-in_, so this would be cumbersome for most. I personally don't think this is the best idea, I just put it here for the sake of completeness.

----

Happy to hear your thoughts on this @molefrog :)

jeetiss:
Patch history once

related to #346
closes #167

### DIFF
--- a/packages/wouter/src/use-browser-location.js
+++ b/packages/wouter/src/use-browser-location.js
@@ -54,13 +54,13 @@ export const useBrowserLocation = (opts = {}) => [
   useEvent((to, navOpts) => navigate(absolutePath(to, opts.base), navOpts)),
 ];
 
+const patchKey = Symbol.for("wouter_v3");
+
 // While History API does have `popstate` event, the only
 // proper way to listen to changes via `push/replaceState`
 // is to monkey-patch these methods.
 //
 // See https://stackoverflow.com/a/4585031
-// eslint-disable-next-line no-undef
-const patchKey = Symbol.for("wouter_v3");
 if (typeof history !== "undefined" && typeof window[patchKey] === "undefined") {
   // patch history object only once
   // See: https://github.com/molefrog/wouter/issues/167
--- a/packages/wouter/src/use-browser-location.js
+++ b/packages/wouter/src/use-browser-location.js
@@ -62,10 +62,6 @@ const patchKey = Symbol.for("wouter_v3");
 //
 // See https://stackoverflow.com/a/4585031
 if (typeof history !== "undefined" && typeof window[patchKey] === "undefined") {
-  // patch history object only once
-  // See: https://github.com/molefrog/wouter/issues/167
-  Object.defineProperty(window, patchKey, { value: true });
-
   for (const type of [eventPushState, eventReplaceState]) {
     const original = history[type];
     // TODO: we should be using unstable_batchedUpdates to avoid multiple re-renders,
@@ -80,4 +76,8 @@ if (typeof history !== "undefined" && typeof window[patchKey] === "undefined") {
       return result;
     };
   }
+
+  // patch history object only once
+  // See: https://github.com/molefrog/wouter/issues/167
+  Object.defineProperty(window, patchKey, { value: true });
 }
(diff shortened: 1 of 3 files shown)


molefrog (changes requested):

packages/wouter/src/use-browser-location.js
```diff
@@ -54,12 +54,18 @@ export const useBrowserLocation = (opts = {}) => [
...
 // While History API does have `popstate` event, the only
 // proper way to listen to changes via `push/replaceState`
 // is to monkey-patch these methods.
 //
 // See https://stackoverflow.com/a/4585031
-if (typeof history !== "undefined") {
+if (typeof history !== "undefined" && typeof window[patchKey] === "undefined") {
+  // patch history object only once
+  // See: https://github.com/molefrog/wouter/issues/167
+  Object.defineProperty(window, patchKey, { value: true });
```
molefrog, line 67: Not a big deal, but shouldn't this property be set after the actual patch? I mean if the code below somehow fails (exception) the patch will not be complete.

jeetiss: yes, it should be nice

packages/wouter/src/use-browser-location.js
```diff
@@ -54,12 +54,18 @@ export const useBrowserLocation = (opts = {}) => [
...
 
+const patchKey = Symbol.for("wouter_v3");
+
 // While History API does have `popstate` event, the only
 // proper way to listen to changes via `push/replaceState`
 // is to monkey-patch these methods.
 //
 // See https://stackoverflow.com/a/4585031
-if (typeof history !== "undefined") {
+if (typeof history !== "undefined" && typeof window[patchKey] === "undefined") {
```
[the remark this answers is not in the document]
jeetiss: This check works in node! `typeof history !== "undefined"` stops the falling
I wanna improve this in the next RP that would be all about `globalThis`. Can it be kept as it is here?

molefrog (approved):

PR #352, opened 2023-10-04
v3: make all hooks from `use-browser-location` compatible with node

### LINKED ISSUE #345: `useSearch` does not work in SSR
`ReferenceError: location is not defined` when doing SSR

```tsx
import { renderToString } from "react-dom/server";
import { Router } from "wouter";
import { App } from "./app";

console.log(
  renderToString(
    <Router ssrPath="/">
      <App />
    </Router>
  )
);
```

```tsx
// app.tsx

import { Link, Route } from "wouter";

export const App = () => (
  <>
    <Route
      path="/"
      component={() => (
        <div>
          <h1>Home</h1>
          <Link to="/about">About</Link>
        </div>
      )}
    />
    <Route
      path="/about"
      component={() => (
        <div>
          <h1>About</h1>
          <Link to="/">Home</Link>
        </div>
      )}
    />
  </>
);
```

jeetiss:
- test all hooks in node

this should fix #345 and #344

### DIFF
--- a/packages/wouter/src/use-browser-location.js
+++ b/packages/wouter/src/use-browser-location.js
@@ -30,7 +30,12 @@ export const useLocationProperty = (fn, ssrFn) =>
   useSyncExternalStore(subscribeToLocationUpdates, fn, ssrFn);
 
 const currentSearch = () => location.search;
-export const useSearch = () => useLocationProperty(currentSearch);
+
+export const useSearch = ({ ssrSearch } = {}) =>
+  useLocationProperty(
+    currentSearch,
+    ssrSearch ? () => ssrSearch : currentSearch
+  );
 
 const currentPathname = () => location.pathname;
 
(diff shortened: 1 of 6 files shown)


packages/wouter/src/use-browser-location.js
```diff
@@ -30,7 +30,12 @@ export const useLocationProperty = (fn, ssrFn) =>
   useSyncExternalStore(subscribeToLocationUpdates, fn, ssrFn);
 
 const currentSearch = () => location.search;
-export const useSearch = () => useLocationProperty(currentSearch);
+
+export const useSearch = ({ ssrSearch } = {}) =>
+  useLocationProperty(
+    currentSearch,
+    ssrSearch ? () => ssrSearch : currentSearch
+  );
```
molefrog, line 38: I've been thinking, what are the cases when we would want to use `currentSearch` on a server? Maybe it's better to have default value of `""` for `ssrSearch`?

If a developer for some reason forgets to set `ssrSearch` while rendering, they will receive an error now, while they might instead render the app with empty params and then receive a hydration error on the client.

molefrog (approved):
🔥

PR #357, opened 2023-10-09
`wouter/use-hash-location`

molefrog:
Fully functional hash location hook with base path support. Also, it can accept `state` when navigating (similarly to React Router). In order to change the state, I had to use `pushState` that keeps pathname and search untouched.

This hook also guarantees to react only to hash changes, e.g. when the search string is modified it won't rerender. Navigation isn't sensitive to leading slash, because, unlike `pathname` it can be missing.
 
- [x] support `state`
- [x] types

### DIFF
--- a/packages/wouter/src/use-hash-location.js
+++ b/packages/wouter/src/use-hash-location.js
@@ -1,7 +1,8 @@
 import { useSyncExternalStore, useEvent } from "./react-deps.js";
 import { absolutePath, relativePath } from "./paths.js";
 
-// fortunately `hashchange` is a native event, so there is need to patch `history` object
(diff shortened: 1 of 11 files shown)


packages/wouter/src/use-hash-location.js
```diff
@@ -0,0 +1,28 @@
+import { useSyncExternalStore, useEvent } from "./react-deps.js";
+import { absolutePath, relativePath } from "./paths.js";
+
+// fortunately `hashchange` is a native event, so there is need to patch `history` object
```
molefrog, line 4:
```suggestion
// fortunately `hashchange` is a native event, so there is no need to patch `history` object
```

jeetiss (approved):

Issue #355, opened 2023-10-07
V3: `memory-location`


molefrog:
The second essential hook that will be included in the next release. Why: mostly for testing and maybe for environments that don't support location (Electron?).

Specs:

```js
// no `use-` here since it is a high order hook
import { memoryLocation } from "wouter/memory-location"

// simple usage
const { hook } = memoryLocation({ path: "/initial-path" })
const [location, navigate] = hook()

// initial path is "/" by default
const { hook } = memoryLocation()

// must support `base` option for nested routing
const [location, navigate] = hook({ base: "/app" })

// standalone `navigate` method for navigating outside of components
const { navigate, hook } = memoryLocation()
navigate("/users")

// `static` option makes it immutable
const { hook } = memoryLocation({ static: true })
navigate("/users") // nothing happens

// `record` option for saving the navigation history
const { hook, navigate, history } = memoryLocation({ record: true })
navigate("/lol")
history // ["/lol", "/"]
```

After this hook is ready, we can migrate our tests to use memory location instead of relying on `"test-utils"`
More ideas?

molefrog:
Closed via #358

PR #358, opened 2023-10-09
v3: memory location

jeetiss:
implement #355

### DIFF
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -1,0 +1,1 @@
+export const memoryLocation = () => ({});
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -1,1 +1,52 @@
-export const memoryLocation = () => ({});
+import mitt from "mitt";
+import { useSyncExternalStore } from "./react-deps.js";
+import { relativePath } from "./paths.js";
+
+/**
+ * In-memory location that supports navigation
+ */
+
+export const memoryLocation = ({
+  path = "/",
+  static: staticLocation,
+  record,
+} = {}) => {
+  let currentPath = path;
+  const history = [];
+  const emitter = mitt();
+
+  const navigateImplementation = (path, { replace = false } = {}) => {
+    if (record) {
+      if (replace) {
+        history.splice(history.length - 1, 1, path);
+      } else {
+        history.push(path);
+      }
+    }
+
+    currentPath = path;
+    emitter.emit("navigate", path);
+  };
+
+  const navigate = !staticLocation ? navigateImplementation : () => null;
+
+  const useMemoryLocation = ({ base } = {}) => {
+    const location = useSyncExternalStore(
+      (cb) => {
+        emitter.on("navigate", cb);
+        return () => emitter.off("navigate", cb);
+      },
+      () => currentPath
+    );
+
+    return [relativePath(base, location), navigate];
+  };
+
+  return {
+    hook: useMemoryLocation,
+    navigate,
+    history: record ? history : undefined,
+  };
+};
+
+// TYPE: { hook: BaseLocationHook; navigate: (path: Path) => void }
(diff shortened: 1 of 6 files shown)


packages/wouter/src/memory-location.js
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  const navigate = !staticLocation ? navigateImplementation : () => null;
+
+  const useMemoryLocation = ({ base } = {}) => {
+    const location = useSyncExternalStore(
+      (cb) => {
+        emitter.on("navigate", cb);
+        return () => emitter.off("navigate", cb);
+      },
```
molefrog, line 38: Oh, that's a common caveat when working with `uSES`, I had this problem once too. React will resubscribe every time the first argument changes, so the reference to this function should stay the same to avoid extra work.

> The store snapshot returned by getSnapshot must be immutable. If the underlying store has mutable data, return a new immutable snapshot if the data has changed. Otherwise, return a cached last snapshot.

packages/wouter/test/memory-location.test-d.ts
```diff
@@ -0,0 +1,43 @@
...
+
+it("should return hook that supports location spec", () => {
+  const { hook } = memoryLocation();
+
+  assertType<Function>(hook);
+
+  const [location, navigate] = hook();
+
+  assertType<string>(location);
+  assertType<Function>(navigate);
```
molefrog, line 12: You can simply use `expectTypeOf(hook).toMatchTypeOf<BaseLocationHook>()` here

packages/wouter/test/memory-location.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  const { result, unmount } = renderHook(() => hook());
+
+  act(() => result.current[1]("/location"));
+
+  const [value] = result.current;
+  expect(value).toBe("/location");
+  unmount();
+});
+
+it("should record all history", () => {
```
molefrog, line 70:
```suggestion
it("should record all history when `record` option is provided", () => {
```

packages/wouter/test/memory-location.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  act(() => standalone("/standalone", { replace: true }));
+
+  expect(history).toStrictEqual(["/test", "/standalone", "/standalone"]);
+
+  act(() => result.current[1]("/location", { replace: true }));
+
+  expect(history).toStrictEqual(["/test", "/standalone", "/location"]);
+
+  unmount();
+});
```
molefrog, line 95: Not super important, but perhaps we could also assert that `history` is undefined when it is not recording?

packages/wouter/test/redirect.test.tsx
```diff
@@ -1,8 +1,20 @@
 import { it, expect } from "vitest";
 import { render } from "@testing-library/react";
+import { useState } from "react";
 
 import { Redirect, Router } from "wouter";
-import { customHookWithReturn } from "./test-utils.js";
+
+export const customHookWithReturn =
```
molefrog, line 7: Thanks, much better!

molefrog (changes requested):

packages/wouter/src/memory-location.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const subscribe = (cb) => {
+    emitter.on("navigate", cb);
+    return () => emitter.off("navigate", cb);
+  };
+
+  const useMemoryLocation = ({ base } = {}) => {
+    const location = useSyncExternalStore(subscribe, () => currentPath);
+
+    return [relativePath(base, location), navigate];
```
molefrog, line 41: I suppose `navigate` method should transform the path provided using `absolutePath` similarly to browser location. How to reproduce:
1. Use base path `"/app"`, initial path `"/app/users"`. The value returned from the hook is `"/users"`.
2. Navigate to `"/settings"`, the path is `"/settings"` (with current setup it will be `"~/settings"`

jeetiss: ready!

molefrog (approved):
🍾

PR #360, opened 2023-10-11
Fix: index.d.ts needs to import Params

tehwalris:
This PR adds a missing import to `index.d.ts`. Without this fix TypeScript will give the following error when you have wouter as a dependency:

```
node_modules/.pnpm/wouter@2.12.0_react@18.2.0/node_modules/wouter/types/ts4.1/index.d.ts:169:71 - error TS2304: Cannot find name 'Params'.

169 export function useParams<T extends DefaultParams = DefaultParams>(): Params<T>;
```

### DIFF
--- a/types/ts4.1/index.d.ts
+++ b/types/ts4.1/index.d.ts
@@ -20,7 +20,7 @@ import {
   LocationHook,
 } from "../use-location";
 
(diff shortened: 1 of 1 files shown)


molefrog (approved):
Thank you!

molefrog:
Published in 2.12.1

PR #364, opened 2023-10-16
v3: Add `reset` method to memory location constructor

jeetiss:
- to drop all history

### DIFF
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -42,11 +42,8 @@ export const memoryLocation = ({
   function reset() {
     // clean history array with mutation to preserve link
     history.splice(0, history.length);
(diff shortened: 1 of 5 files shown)


packages/wouter/src/memory-location.js
```diff
@@ -39,9 +39,15 @@ export const memoryLocation = ({
     navigate,
   ];
 
+  function reset() {
+    history.splice(0, history.length);
+    history.push(path);
```
molefrog, line 44: Shouldn’t reset trigger the update event and set the location? Imo, when called while the app is still mounted we should rerender everything to be consistent with the new location

molefrog (approved):

molefrog (approved):

PR #373, opened 2023-10-25
`useSearch` is part of `"wouter"`

molefrog:
This PR makes it possible to import `useSearch` directly from `"wouter"`. By default, it is the same as browser search hook from `"use-browser-location"`, however it can be customized via `<Router searchHook={custom} />`


jeetiss (approved):
🤩

PR #375, opened 2023-11-01
Ship `regexparam` with optional wildcards

molefrog:
Adds `src/regexparam.js` which was extracted from https://github.com/lukeed/regexparam
It is a temporary fix until https://github.com/lukeed/regexparam/pull/25 is shipped


molefrog:
@jeetiss I've decided not to create a workspace package, since it makes it harder to maintain builds and ship types. The code itself is 30 LOC, so I guess it is fine for the time being. I really hope that eventually it lands there, so we could remove the fix.

jeetiss (approved):

PR #378, opened 2023-11-05
v3: link with as child pattern

jeetiss:
`<Link asChild />` feature

### usage

```
// right

<Link to='href' ref={ref}>
  link
</Link>

<Link to='href' ref={ref} className={'hello'}>
  link
</Link>

<Link to='href' asChild>
  <Button>hello world</Button>
</Link>

<Link to='href' asChild>
  <Button ref={ref}>hello world</Button>
</Link>

// wrong

<Link to='href' asChild ref={ref} className={'noo'}>
  <Button ref={ref}>hello world</Button>
</Link>

<Link to='href' asChild>
  text
</Link>
```

### DIFF
--- a/packages/wouter-preact/types/index.d.ts
+++ b/packages/wouter-preact/types/index.d.ts
@@ -81,9 +81,9 @@ export type NavigationalProps<
 > = ({ to: Path; href?: never } | { href: Path; to?: never }) &
   HookNavigationOptions<H>;
 
-type AsChildProps<RequiredProps, DefaultElementProps> =
-  | ({ asChild?: false } & RequiredProps & DefaultElementProps)
-  | ({ asChild: true; children: ComponentChildren } & RequiredProps);
+type AsChildProps<ComponentProps, DefaultElementProps> =
+  | ({ asChild?: false } & ComponentProps & DefaultElementProps)
+  | ({ asChild: true; children: ComponentChildren } & ComponentProps);
 
 export type LinkProps<H extends BaseLocationHook = BrowserLocationHook> =
   AsChildProps<NavigationalProps<H>, JSX.HTMLAttributes>;
--- a/packages/wouter-preact/types/index.d.ts
+++ b/packages/wouter-preact/types/index.d.ts
@@ -81,11 +81,12 @@ export type NavigationalProps<
 > = ({ to: Path; href?: never } | { href: Path; to?: never }) &
   HookNavigationOptions<H>;
 
-export type LinkProps<H extends BaseLocationHook = BrowserLocationHook> = Omit<
-  JSX.HTMLAttributes,
-  "href"
-> &
-  NavigationalProps<H>;
+type AsChildProps<RequiredProps, DefaultElementProps> =
+  | ({ asChild?: false } & RequiredProps & DefaultElementProps)
+  | ({ asChild: true; children: ComponentChildren } & RequiredProps);
+
+export type LinkProps<H extends BaseLocationHook = BrowserLocationHook> =
+  AsChildProps<NavigationalProps<H>, JSX.HTMLAttributes>;
 
 export type RedirectProps<H extends BaseLocationHook = BrowserLocationHook> =
   NavigationalProps<H> & {
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -189,8 +189,10 @@ export const Link = forwardRef((props, ref) => {
     onClick: _onClick,
     asChild,
     children,
+    /* eslint-disable no-unused-vars */
     replace /* ignore nav props */,
     state /* ignore nav props */,
+    /* eslint-enable no-unused-vars */
     ...restProps
   } = props;
 
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -215,7 +215,15 @@ export const Link = forwardRef((props, ref) => {
     });
   }
 
-  return h("a", { ...props, href: link, onClick: handleClick, to: null, ref });
+  return h("a", {
+    ...props,
+    href: link,
+    onClick: handleClick,
+    to: null,
+    replace: null,
+    state: null,
+    ref,
+  });
 });
 
 const flattenChildren = (children) => {
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -204,12 +204,15 @@ export const Link = forwardRef((props, ref) => {
     }
   });
 
+  // handle nested routers and absolute paths
   const link = href[0] === "~" ? href.slice(1) : router.base + href;
 
   if (asChild) {
-    const jsx = isValidElement(children) ? children : null;
-    return cloneElement(jsx, {
-      // handle nested routers and absolute paths
+    if (!isValidElement(children)) {
+      throw Error("Only one child allowed");
+    }
+
+    return cloneElement(children, {
       href: link,
       onClick: handleClick,
     });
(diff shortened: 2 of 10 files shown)


packages/wouter/src/index.js
```diff
@@ -204,17 +204,29 @@ export const Link = forwardRef((props, ref) => {
...
-  // wraps children in `a` if needed
-  const extraProps = {
-    // handle nested routers and absolute paths
-    href: href[0] === "~" ? href.slice(1) : router.base + href,
+  // handle nested routers and absolute paths
+  const link = href[0] === "~" ? href.slice(1) : router.base + href;
+
+  if (asChild) {
+    if (!isValidElement(children)) {
+      throw Error("Only one child allowed");
```
jeetiss, line 212: I'm not sure about `error` here. is it okay to trow error when children are invalid?

@molefrog what do you think?

packages/wouter-preact/types/index.d.ts
```diff
@@ -81,11 +81,12 @@ export type NavigationalProps<
...
-  JSX.HTMLAttributes,
-  "href"
-> &
-  NavigationalProps<H>;
+type AsChildProps<RequiredProps, DefaultElementProps> =
+  | ({ asChild?: false } & RequiredProps & DefaultElementProps)
+  | ({ asChild: true; children: ComponentChildren } & RequiredProps);
+
+export type LinkProps<H extends BaseLocationHook = BrowserLocationHook> =
+  AsChildProps<NavigationalProps<H>, JSX.HTMLAttributes>;
```
jeetiss, line 89: I have no experience with preact types, could you review it with more passion?

packages/wouter/test/link.test-d.tsx
```diff
@@ -1,44 +1,60 @@
...
-      <Link href="/" state={undefined}>
-        test
-      </Link>
-    );
-    assertType(
-      <Link href="/" state="string">
-        test
-      </Link>
-    );
+  it("should throw error when `to` and `href` props are used in same time", () => {
```
molefrog, line 12: I would rephrase this, because technically it doesn't throw an error. Maybe:
```suggestion
  it("does not allow `to` and `href` props to be used at the same time", () => {
```

packages/wouter/test/link.test-d.tsx
```diff
@@ -1,44 +1,60 @@
...
+    // @ts-expect-error
+    <Link to="/hello" href="/world">
+      Hello
+    </Link>;
+  });
+
+  it("should inherit props from `HTMLAnchorElements`", () => {
```
molefrog, line 19:
```suggestion
  it("should inherit props from `HTMLAnchorElement`", () => {
```

packages/wouter/test/link.test-d.tsx
```diff
@@ -56,19 +72,61 @@ describe("Link's ref", () => {
...
-    <Link to="/">
-      <Component ref={ref}></Component>
+  it("should throw error when `to` and `href` props are used in same time", () => {
+    // @ts-expect-error
+    <Link to="/hello" href="/world" asChild>
+      <a>Hello</a>
+    </Link>;
+  });
+
+  it("should throw error when unsupported props are provided", () => {
```
molefrog, line 94: Same here.

packages/wouter/test/link.test-d.tsx
```diff
@@ -56,19 +72,61 @@ describe("Link's ref", () => {
...
+    // @ts-expect-error
+    <Link to="/" asChild ref={null}>
+      <a>Hello</a>
     </Link>;
   });
-});
 
-const Component = React.forwardRef<{ hello: "world" }>((props, ref) => {
-  return <>test</>;
+  it("should throw error when to and href props are used in same time", () => {
```
molefrog, line 116: And here.

packages/wouter/test/link.test-d.tsx
```diff
@@ -56,19 +72,61 @@ describe("Link's ref", () => {
...
   });
+});
 
-  it("should work with composed components", () => {
-    const ref = React.useRef<React.ElementRef<typeof Component>>(null);
+describe("<Link /> in asChild mode", () => {
+  it("should work", () => {
+    <Link to="/" asChild>
+      <a>Hello</a>
+    </Link>;
```
molefrog, line 84: I would also test custom component just in case to show that not only links are allowed here.

packages/wouter/test/link.test-d.tsx
```diff
@@ -1,44 +1,60 @@
...
+  });
+
+  it("should support other navigation params", () => {
+    <Link href="/" state={{ a: "foo" }}>
+      test
+    </Link>;
+
+    <Link href="/" replace>
+      test
+    </Link>;
```
molefrog, line 44: At the same time, using anything other than boolean as `replace` should not be allowed. Could you test that as well?

packages/wouter/test/link.test-d.tsx
```diff
@@ -1,44 +1,60 @@
...
+
+    <Link href="/" state={undefined}>
+      test
+    </Link>;
   });
```
molefrog, line 49: Imo, we should also test generic Link type with custom navigation hook. Let's say for example `useNetworkLocation` and `navigate({ host: "whatever", retries: 4 })`.

packages/wouter/test/link.test.tsx
```diff
@@ -1,206 +1,221 @@
...
+  });
 });
 
-it("renders `href` with basepath", () => {
-  const { getByTestId } = render(
-    <Router base="/app">
-      <Link href="/dashboard" data-testid="link" />
-    </Router>
-  );
+describe("<Link /> in asChild mode", () => {
```
molefrog, line 169:
```suggestion
describe("<Link /> with `asChild` props", () => {
```

packages/wouter/test/link.test.tsx
```diff
@@ -1,206 +1,221 @@
...
+  it("works for any other elements as well", () => {
```
molefrog, line 170: I think the test case label is a bit confusing. How about:

```suggestion
  it("when `asChild` is not specified, wraps the children in an <a />", () => {
```

packages/wouter/types/index.d.ts
```diff
@@ -113,8 +107,18 @@ export function Redirect<H extends BaseLocationHook = BrowserLocationHook>(
   context?: any
 ): ReactElement<any, any> | null;
 
+type AsChildProps<RequiredProps, DefaultElementProps> =
```
molefrog, line 110: What do you think about renaming `RequiredProps` to something like `NavigateOptions`?

molefrog (changes requested):

packages/wouter-preact/types/index.d.ts
molefrog: I am not an expert either but could you try `JSX.IntrinsicElements["a"]`? This should give attributes tailored to `a` (mostly to get the proper type in event handlers).

packages/wouter/types/index.d.ts
jeetiss: This is a generic type that can be used with other components, not only with links.

I renamed this prop to `ComponentProps`, is this a better name?

packages/wouter/test/link.test-d.tsx
jeetiss: typescript can't check children types, so if a regular one works, custom would work too

packages/wouter-preact/types/index.d.ts
jeetiss: `JSX.IntrinsicElements["a"]` nope
it doesn't work

molefrog (approved):
👍

packages/wouter-preact/types/index.d.ts
molefrog: Ok!

molefrog (approved):

PR #388, opened 2023-11-16
Allow links to accept `onClick` when used in `asChild` mode

molefrog:
Also, I narrowed the type of children to be a valid React element.


jeetiss (approved):
👍🏻

PR #391, opened 2023-11-30
Feat: `useSearchParams` utility hook

### LINKED ISSUE #368: Need support for search params.
Need something like useSearchParams in React Router to handle queries.

junwen-k:
This PR introduce a `useSearchParams` utility hook that is similar to React-Router's [`useSearchParams`](https://reactrouter.com/en/main/hooks/use-search-params) hook.

Due to the minimalistic approach of this library, the utility hook is designed to be as simple as possible while still being useful for most cases. We could even remove the callback `setState` style if want to be truly minimalist. It does not support "default" search params, does not support array values out of the box to keep the API surface minimal. Let me know what do you think :)

It respects `useSearch` and `useLocation` and uses them internally for the utility hook.

This is based on my understanding with this library, feel free to give any feedbacks :)

Closes #368

### DIFF
--- a/packages/wouter/src/use-browser-location.js
+++ b/packages/wouter/src/use-browser-location.js
@@ -1,4 +1,9 @@
-import { useCallback, useMemo, useSyncExternalStore } from "./react-deps.js";
+import {
+  useRef,
+  useCallback,
+  useMemo,
+  useSyncExternalStore,
+} from "./react-deps.js";
 
 /**
  * History API docs @see https://developer.mozilla.org/en-US/docs/Web/API/History
@@ -35,19 +40,22 @@ export const useSearch = ({ ssrSearch = "" } = {}) =>
 
 export const useSearchParams = ({ ssrSearch = "" } = {}) => {
   const search = useSearch({ ssrSearch });
-  const searchParams = useMemo(() => new URLSearchParams(search), [search]);
-
-  const setSearchParams = useCallback(
-    (nextInit, navOpts) => {
-      const newSearchParams = new URLSearchParams(
-        typeof nextInit === "function" ? nextInit(searchParams) : nextInit
-      );
-      navigate("?" + newSearchParams, navOpts);
-    },
-    [searchParams]
+  const searchParamsRef = useRef(new URLSearchParams(search));
+  searchParamsRef.current = useMemo(
+    () => new URLSearchParams(search),
+    [search]
   );
 
-  return [searchParams, setSearchParams];
+  const setSearchParams = useCallback((nextInit, navOpts) => {
+    const newSearchParams = new URLSearchParams(
+      typeof nextInit === "function"
+        ? nextInit(searchParamsRef.current)
+        : nextInit
+    );
+    navigate("?" + newSearchParams, navOpts);
+  }, []);
+
+  return [searchParamsRef.current, setSearchParams];
 };
 
 const currentPathname = () => location.pathname;
--- a/packages/wouter/src/use-browser-location.js
+++ b/packages/wouter/src/use-browser-location.js
@@ -1,9 +1,4 @@
-import {
-  useRef,
-  useCallback,
-  useMemo,
-  useSyncExternalStore,
-} from "./react-deps.js";
+import { useEvent, useSyncExternalStore } from "./react-deps.js";
 
 /**
  * History API docs @see https://developer.mozilla.org/en-US/docs/Web/API/History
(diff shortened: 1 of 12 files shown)


junwen-k:
I have yet to include any documentation in the README though. But feel free to take a look at the implementation details first before we proceed to documentation.

packages/wouter/src/use-browser-location.js
```diff
@@ -33,6 +38,26 @@ const currentSearch = () => location.search;
...
+      typeof nextInit === "function"
+        ? nextInit(searchParamsRef.current)
+        : nextInit
+    );
+    navigate("?" + newSearchParams, navOpts);
+  }, []);
+
+  return [searchParamsRef.current, setSearchParams];
+};
+
```
molefrog, line 60: Just some ideas to reduce extra React imports and make this function smaller:

```suggestion
export const useSearchParams = ({ ssrSearch = "" } = {}) => {
  const search = useSearch({ ssrSearch });
  const params = useMemo(() => new URLSearchParams(search)));
  
  const setSearchParams = useEvent((nextInit, navOpts) => {
    const newSearchParams = new URLSearchParams(
      typeof nextInit === "function"
        ? nextInit(params)
        : nextInit
    );
    
    navigate("?" + newSearchParams, navOpts);
  });

  return [params, setSearchParams];
};
```

Here, I replaced `useCallback` with `useEvent`. We should not see degraded performance, unless the component that uses it renders too often, e.g. once every 50-300 ms (but that isn't really our responsibility, since when used standalone this hook will re-render only when the query string changes, which in real-life scenarios shouldn't be too often). Plus, `useEvent` is already part of `react-deps.js`.

I'm still trying to figure out how `useMemo` can be avoided further on. The question is whether we really want to memoise it. Considering, that the hook doesn't render too often, it should be safe to always return a new instance of `URLSearchParams`. Why:
1. **Immutability.** Is it safe to re-use this object? What if developer decides to mutate it for some reason, this would cause some undefined bevahiour. 
2. **The memory/CPU footprint of `URLSearchParams`.**  I'm not sure how browsers implement it, but at the first glance it doesn't seem like a heavy resource. Though, I will make a proper research to figure this out.
3. **Stability.** One of the caveats could be that we always return a new object reference. This could harm performance only if `searchParams` instance is further passed down to components as a prop. But in reality, why would someone want to do that? I assume that the normal use-case is the get the object, and then extract the required primitive objects from it, e.g. `searchParams.get("foo")`

Anyways, there is a trade-off that we should evaluate there.

junwen-k:
I think I will make the changes required after resolving https://github.com/molefrog/wouter/issues/393 first.
I might also remove the `useMemo` as well.

Thanks @molefrog

junwen-k:
@molefrog Hi, any chance to see `searchParams` officially supported in this library? Related PR https://github.com/molefrog/wouter/pull/422 :)

MatejFacko:
Hello guys, any update on this? we would love to use this hook :)

theguacamoleking:
Hello! Lurker here with some info that might be of use.

I tried implementing something similar to this in the past where search param changes cause a navigation event, triggering a rerender from which the new state can be inferred from the history API. While this worked on Chrome, I found that Safari will throw an error if more than 100 history pushes occur within 30s

> SecurityError: Attempt to use history.pushState() more than 100 times per 10 seconds

You can trigger this pretty easily in the console 

```js
setInterval(() => history.pushState(undefined, undefined, location.href.replace(/\?.*/, "") + `?test=${Math.random()}`),10)
```

So yeah, just a heads up as this limitation isn't super well known and can fundamentally break an application in cases where application state is coupled to query params.

From what I can tell, Wouter is using the history API as the main source of truth for state - I assume other libraries avoid this issue by decoupling application state from the history API (and debouncing history pushes)?

not-first:
Hey, I was wondering if there was any updates whether this would be included in the library. I would love to use the hook!

guoyunhe:
Since React Router v7 is a real breaking change, we are planning to migrate to Wouter. However, we do need a `useSearchParams` hook, because we used it heavily in our code base. Would be great to see any progress in this PR.

@junwen-k before this PR is merged, is it possible that you can publish the hook a standalone package, e.g. wouter-search ? People can help try and test it.

junwen-k:
> Since React Router v7 is a real breaking change, we are planning to migrate to Wouter. However, we do need a `useSearchParams` hook, because we used it heavily in our code base. Would be great to see any progress in this PR.
> 
> @junwen-k before this PR is merged, is it possible that you can publish the hook a standalone package, e.g. wouter-search ? People can help try and test it.

Hi @guoyunhe,

Thanks for your interest and for bringing this up! Since working on this PR, I’ve actually moved away from Wouter to other routing solutions like [TanStack Router](https://tanstack.com/router/latest/docs/framework/react/overview) and [Next.js](https://nextjs.org/). I’ve found that I prefer the declarative, file-based routing approach (similar to Next.js) for its clarity and maintainability. TanStack Router's [file-based routing](https://tanstack.com/router/latest/docs/framework/react/guide/file-based-routing) provides a similar developer experience to Next.js, but without requiring Next.js itself.  

If you’re exploring alternatives, you might want to check out [nuqs](https://github.com/47ng/nuqs). It has some interesting features for managing search params and offers first-class support for different routing solutions through adapters. I’ve tried it with Next.js, and it’s been working pretty well so far!  

Unfortunately, I won’t be publishing this as a standalone package, as I’ve transitioned to these newer solutions. That said, I’d be happy to support or guide anyone interested in taking over this work.  

Thanks again! :)

guoyunhe:
@junwen-k FYI, I created a package wouter-search based on your PR :christmas_tree: https://github.com/guoyunhe/wouter-search

molefrog:
Closing in favour of #503

PR #411, opened 2024-02-02
Routes with no path fall back to "*"

molefrog:
A small sugar to save up some space, make the code more elegant and enable some nice DX tricks. This is the last commit before the V3 release!


jeetiss (approved):
💅

PR #426, opened 2024-03-13
Fix: add nest prop to Preact types, in line with React types

bensaufley:
Am I wrong that this is just an oversight in the types? The Preact package mostly just imports the React one. I did double-check the rest of the file and everything looks equivalent besides this.


molefrog:
You are right, I've overlooked it. Preact types sometimes become out of date because we lack a good way to build them based on React types. This probably requires some refactoring and a Vite plugin (haven't researched this, but open to contributions).

molefrog (approved):

molefrog:
Published a bugfix in v3.1.1

PR #428, opened 2024-03-25
Fix compilation error on React < 18

alex-pex:
The error was `Attempted import error: 'useInsertionEffect' is not exported from 'react' (imported as 'React').`

I have used the same technique as Emotion https://github.com/emotion-js/emotion/commit/75a2f74418019819e0cdcb1e0720b0a5d0d65687#diff-e87fa5e3ae2595fc1671477ceb9c7327f968d5556791d29958210ed43474fc92R17

### DIFF
--- a/packages/wouter/src/react-deps.js
+++ b/packages/wouter/src/react-deps.js
@@ -1,12 +1,14 @@
 import * as React from "react";
 
-// React.useInsertionEffect is not available in React <18
 const {
   useEffect,
   useLayoutEffect,
   useRef,
-  useInsertionEffect: useBuiltinInsertionEffect,
 } = React;
+
+// React.useInsertionEffect is not available in React <18
+// This hack fixes a transpilation issue on some apps
+const useBuiltinInsertionEffect = React['useInsertion' + 'Effect'];
 
 export {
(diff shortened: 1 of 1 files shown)


alex-pex:
I don't know if you really need the bulletproof fix: do people embed wouter inside another package and still expect it to work with react < 18? It's a strange combo

molefrog:
I see, could you also add a comment above that line to explain the hack?

molefrog:
Also, one thing I can't wrap my head around is the import of `useSyncExternalStore` (also, React 18 only) in the shim library we use. They are essentially doing the same thing: [importing the native method](https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShim.js#L13) with a fallback to the user-land implementation. Wouter depends on this library, so I would expect webpack to complain about the import there too. Can you try to import this library directly for the sake of experiment?

alex-pex:
image
with `"use-sync-external-store": "^1.2.0",`

molefrog:
Released in 3.1.2

alex-pex:
thanks!

Issue #435, opened 2024-04-09
build bug


huazhuangnan:
RollupError: [commonjs--resolver] node_modules/wouter/esm/react-deps.js (8:2): Identifier "useRef" has already been declared

molefrog:
I can't confirm this error, can you provide more details? As I can see, in `wouter/esm/react-deps.js` `useRef` is only declared once.

huazhuangnan:
Yes, Rollup builds will fail!

PR #441, opened 2024-04-24
chore: tweak imports

jeetiss:
remove destructuring assignment for react wildcard import because it prevents tree-shaking and removing makes code little bit smaller

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6499,7 +6499,7 @@
       }
     },
     "packages/wouter": {
-      "version": "3.1.0",
+      "version": "3.1.2",
       "license": "Unlicense",
       "dependencies": {
         "mitt": "^3.0.1",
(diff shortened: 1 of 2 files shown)


jeetiss:
I guess this should fix https://github.com/molefrog/wouter/issues/435 as well

molefrog:
Thank you. In https://github.com/molefrog/wouter/pull/428 we added a hack to ensure that `useInsertionEffect` isn't imported directly. Now we use a dynamic expression, would tree shaking work anyway? Is there any way to test this?

molefrog:
> Thank you. In #428 we added a hack to ensure that `useInsertionEffect` isn't imported directly. Now we use a dynamic expression, would tree shaking work anyway? Is there any way to test this?

@jeetiss What are your thoughts on it?

jeetiss:
@molefrog yep, hack with `useInsertionEffect` is stoping tree-shaking
i think this can be tested with size-limit, we just have to remove react from ignore list

but tree-shakability testing makes no sense to me now
- all applications use all imports from react usually
- it is broken by `useInsertionEffect`^ and we can't fix it

molefrog (approved):
I'm going to merge this anyway since it doesn't break anything, but might also fix the #435

PR #444, opened 2024-05-09
Fix active link highlighting when using custom href formatting

### LINKED ISSUE #417: [BUG] `useHashLocation` not syncing with state
With the `useHashLocation` hook, the active route does not sync with the current state/location, once it is updated. The initial location loads the correct route.

For example/reproduction:
1. Set up a project with wouter and hash routing.
2. Create a route, let's call it `test`. Add a link from it to the index (`#/`) or a `navigate()` call (using `useLocation`)
3. Go to a `#/test`
4. Click the link you added. The location will be updated, but the active route will not change.

neoeno:
_See https://github.com/molefrog/wouter/issues/417#issuecomment-2098847560 for details on the problem._

Previously when we formatted the `href` attribute (for example, to prefix a `#` when using useHashLocation) this would make it not match the current path, which would not have the hash prefix.

This patch preserves the target (href, or to) of the Link component and then compares this value to the active path. This should then match.

I have also altered some variable names to improve clarity given we now have four different various 'paths' in this component.

I have tested this: in this repo, in my previous reproduction, and in the application I am building — so I feel reasonably confident it fixes my issue at least :)

Fixes #417

### DIFF
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -187,11 +187,11 @@ export const Route = ({ path, nest, match, ...renderProps }) => {
 
 export const Link = forwardRef((props, ref) => {
   const router = useRouter();
-  const [path, navigate] = useLocationFromRouter(router);
+  const [currentPath, navigate] = useLocationFromRouter(router);
 
   const {
     to,
-    href: _href = to,
+    href: targetPath = to,
     onClick: _onClick,
     asChild,
     children,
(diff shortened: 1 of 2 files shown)


packages/wouter/src/index.js
```diff
@@ -236,7 +236,7 @@ export const Link = forwardRef((props, ref) => {
         onClick,
         href,
         // `className` can be a function to apply the class if this link is active
-        className: cls?.call ? cls(path === href) : cls,
+        className: cls?.call ? cls(currentPath === (targetPath)) : cls,
```
molefrog: Everything looks good! One small question: do we need extra braces around `targetPath` here?

molefrog (approved):
Looks and works well! Left one minor comment, but other than that it is ready to be merged.

packages/wouter/src/index.js
neoeno: Definitely not :) No idea what I was thinking. Thanks for the catch — fix pushed.

molefrog:
Released in v3.1.3

PR #450, opened 2024-05-28
Added support for regular expressions in paths

### LINKED ISSUE #449: Allow using `RegExp` for paths alongside strings
The `regexparam` library allows for passing `RegExp` through `parse`. I think it would be a useful feature to be able to use custom regular expressions for routes.

For example:
```jsx
<Switch>
  <Route path="/:id">Hello</Route>
  <Route>404</Route>
</Switch>
```

`id` will match anything. But if we know that an ID will only be digits, it would be nice to use a regex to validate it:
```jsx
const idRegex = /^[/](?<id>\d+)$/;

<Switch>
  <Route path={idRegex}>Hello</Route>
  <Route>404</Route>
</Switch>
```
If it doesn't match, it will continue on to the 404 page.

This uses a named capture group so it could be a direct drop in for `"/:id"`.

Alternatively, a normal regex could be used:
```js
/^[/](\d+)$/

// Then consumed like:
const params = useParams();
const id = params[1];
```

This allows for lifting validation out of routes.
Patterns like route prefixes (`/@:username`) and multiple params in a single segment (`/:id-:opt`) becomes possible.
And since it builds upon functionality that is already here, it won't cost much to implement.

JonahPlusPlus:
This fixes #449.

By supporting regular expressions in paths, it becomes possible to lift validation out of routes and into the paths themselves.

What happens when you use a `RegExp`:
The route doesn't run the parser, but just executes the regex. After executing the regex, instead of mapping the keys and values into an object, the indices of the regex result are mapped to an object and the named groups are assigned to the object as well.

For example:
```js
// Route
/[/]([a-z]+)/
// Against "/foo", produces
{ 0: "foo" }

// Route
/[/](?<name>[a-z]+)/
// Against "/foo", produces
{ 0: "foo", name: "foo" }
```

Another thing to consider: should normal paths also expose params as indices? (so, should `/:id` become `{ 0: "foo", id: "foo" }`, instead of just `{ id: "foo" }`?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,9 +221,9 @@ useRoute("/orders/*?");
 
 // regex for matching complex patterns,
 // matches "/hello:123"
-useRoute(/[/]([a-z]+):([0-9]+)[/]?/);
+useRoute(/^[/]([a-z]+):([0-9]+)[/]?$/);
 // and with named capture groups
-useRoute(/[/](?<word>[a-z]+):(?<num>[0-9]+)[/]?/);
+useRoute(/^[/](?<word>[a-z]+):(?<num>[0-9]+)[/]?$/);
 ```
 
 The second item in the pair `params` is an object with parameters or null if there was no match. For wildcard segments the parameter name is `"*"`:
@@ -335,7 +335,7 @@ const User = () => {
   params[0]; // "1"
 };
 
-<Route path={/[/]user[/](?<id>[0-9]+)[/]?/} component={User}> />
+<Route path={/^[/]user[/](?<id>[0-9]+)[/]?$/} component={User}> />
 ```
 
 ### `useSearch`: query strings
@@ -442,9 +442,10 @@ If you call `useLocation()` inside the last route, it will return `/orders` and 
 </Route>
 ```
 
-**Note:** The `nest` prop has no effect on regexes passed in.
-It will only determine if nested routes will match the rest of path or match against the same path.
-To make a strict path regex, use regex techniques like `[/]?$` (this matches an optional end slash and the end of the string).
+**Note:** The `nest` prop does not alter the regex passed into regex paths.
+Instead, the `nest` prop will only determine if nested routes will match against the rest of path or the same path.
+To make a strict path regex, use a regex pattern like `/^[/](your pattern)[/]?$/` (this matches an optional end slash and the end of the string).
+To make a nestable regex, use a regex pattern like `/^[/](your pattern)(?=$|[/])/` (this matches either the end of the string or a slash for future segments).
 
 ### `<Link href={path} />`
 
--- a/README.md
+++ b/README.md
@@ -318,12 +318,15 @@ const User = () => {
   const params = useParams();
 
   params.id; // "1"
+
+  // alternatively, use the index to access the prop
+  params[0]; // "1"
 };
 
 <Route path="/user/:id" component={User}> />
 ```
 
-For regex paths, parameters are accessible as indices or through their group name.
+It is the same for regex paths. Capture groups can be accessed by their index, or if there is a named capture group, that can be used instead.
 
 ```js
 import { Route, useParams } from "wouter";
--- a/README.md
+++ b/README.md
@@ -218,6 +218,12 @@ useRoute("/app*");
 // optional wildcards, matches "/orders", "/orders/"
 // and "/orders/completed/list"
 useRoute("/orders/*?");
+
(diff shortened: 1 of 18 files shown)


JonahPlusPlus:
I changed it to handle the `RegExp` outside of the parser, so devs don't need to handle that case. With that, these changes aren't breaking (AFAIK; it's still a change to the public API).

molefrog:
@JonahPlusPlus Hey, thank you! 👏 You've done an enormous work and everything seems good. I've looked briefly, so I'm going to take some time to understand better how it works.

README.md
```diff
@@ -421,6 +442,10 @@ If you call `useLocation()` inside the last route, it will return `/orders` and
 </Route>
 ```
 
+**Note:** The `nest` prop has no effect on regexes passed in.
```
molefrog: Does that mean nested routes can't work with regexp paths?

packages/wouter/types/router.d.ts
```diff
@@ -5,7 +5,10 @@ import {
   BaseSearchHook,
 } from "./location-hook.js";
 
-export type Parser = (route: Path) => { pattern: RegExp; keys: string[] };
+export type Parser = (
+  route: Path,
+  loose?: boolean
+) => { pattern: RegExp; keys: string[] };
```
molefrog, line 11: Would it be possible to add a quick type spec to cover this change? `router.test-d.tsx`

packages/wouter/types/index.d.ts
```diff
@@ -61,23 +63,33 @@ export interface RouteComponentProps<T extends DefaultParams = DefaultParams> {
...
+        : RegexRouteParams
+    >
   >;
   nest?: boolean;
 }
 
 export function Route<
   T extends DefaultParams | undefined = undefined,
-  RoutePath extends Path = Path
+  RoutePath extends Path | RegExp = Path | RegExp
```
molefrog: Just out of curiosity, if we change the type of `Path` to be `string | RegExp`, will it break anything? It isn't a breaking change since it extends the type.

packages/wouter/src/index.js
```diff
@@ -79,21 +79,45 @@ export const useSearch = () => {
 };
 
 const matchRoute = (parser, route, path, loose) => {
+  // if the input is a regexp, skip parsing
+  const { pattern, keys } = (() => {
+    if (route instanceof RegExp) {
+      return { keys: false, pattern: route };
+    }
+    return parser(route || "*", loose);
+  })();
```
molefrog: Would ternary expression work better here maybe?

README.md
JonahPlusPlus: No, the `nest` prop usually affects the regex generated from string paths. In the case of regex paths, it doesn't alter the given regex.

The `nest` prop still alters the path given to nested routes.

An example (since I'm not sure if I'm wording this correctly):

The path `"/foo"` gets converted to something like `/^[/]foo[/]?$/`.
With the `nest` prop, it becomes something like `/^[/]foo(?=$|[/])/` and removes `/foo` from the path for nested components.

For the path `/[/]foo/`, there is no conversion.
With the `nest` prop, there is also no conversion, but `/foo` is removed from the path for nested components. This regex will likely have issues with nesting (it doesn't determine the start or end), but it is on the developer to make decisions about the regex.

On that note, maybe I should make the regex examples better.

packages/wouter/types/index.d.ts
JonahPlusPlus: I tried that, but some things use `Path` where only a string would make sense (e.g. `NavigationalProps`). I suppose I could add a new `PathPattern` type to represent `string | RegExp`.

JonahPlusPlus:
Okay, so I got those changes done and I added one final change: string paths also return indices for the params as well as the keys.

So, matching `"/:id"` against `/foo` returns `{ 0: "foo", id: "foo" }`.
This is now consistent with regex paths.
Keys override indices, so this should not be a breaking change for any library that uses number keys.
(e.g. something that acts like `"/:1/:0"` won't break order)

Future work:
Provide a way to optimize routes by caching string path regexes.
Possible solution: allow passing `{ pattern, keys }` objects to routes and allow for generating those objects outside of routes.

```js

const idPath = cachePath("/:id");

<Route path={idPath}>
...
</Route>
```

molefrog (approved):
Amazing job! I have no more comments 😃

PR #456, opened 2024-06-17
Export `matchRoute` function

### LINKED ISSUE #455: Consider exposing `matchRoute` helper function
I wanted to ask if you would consider exposing the [`matchRoute`](https://github.com/molefrog/wouter/blob/6abe3832430c0b564ca0054f79ea5e2ff592a1af/packages/wouter/src/index.js#L81) helper function that is internally used by `useRoute` and `Switch`.

I'm trying to create a custom hook which receives a list of routes in order to check which one matches, but this list of routes is dynamic, so it's not possible to call `useRoute` a bunch of times, since it needs to loop over the list of routes.

However, if `matchRoute` was exported by this package, it would allow userland programmatic logic like this:

```ts
import { matchRoute, useLocation, useRouter } from 'wouter-preact';

function useMatchingRoute(candidates: string[]): [string, Record] | undefined {
  const { parser } = useRouter();

  return useMemo(() => {
    for (const route of candidates) {
      const [matches, params] = matchRoute(parser, route, location);
      if (matches) {
        return [route, params];
      }
    }

    return undefined;
  }, [location, parser, candidates]);
}
```

I can provide a PR if you are ok with this.

acelaya:
This PR allows users to import the `matchRoute` function that is internally used by `useRoute` or `<Switch />`, allowing custom dynamic logic or route checks outside of component scope.

Closes https://github.com/molefrog/wouter/issues/455

### DIFF
--- a/packages/wouter/test/match-route.test-d.ts
+++ b/packages/wouter/test/match-route.test-d.ts
@@ -1,0 +1,55 @@
+import { it, expectTypeOf, assertType } from "vitest";
+import { matchRoute, useRouter } from "wouter";
+
+const { parser } = useRouter();
+
+it("should only accept strings", () => {
+  // @ts-expect-error
+  assertType(matchRoute(parser, Symbol(), ""));
+  // @ts-expect-error
+  assertType(matchRoute(parser, undefined, ""));
+  assertType(matchRoute(parser, "/", ""));
+});
+
+it('has a boolean "match" result as a first returned value', () => {
+  const [match] = matchRoute(parser, "/", "");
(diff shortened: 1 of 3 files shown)


packages/wouter/test/match-route.test-d.ts
acelaya, line 1: I added a type test, but I didn't add a regular logic test, as the `useRoute` test already covers all paths inside `matchRoute`.

molefrog (approved):
Looks good! Thank you

Simon-Tang:
This is great :) Just wanted to mention that the documentation doesn't list `matchRoute`, so finding this PR was how I discovered the function.

PR #465, opened 2024-07-03
useHashLocation must trim query string, useSearch must parse query string from location.hash

GravityTwoG:
While using the useHashLocation hook, the current location is retrieved from location.hash, including the query string (everything after "?"). This breaks route matching. 

Additionally, when useHashLocation is used, the useSearch hook works incorrectly. The location.search property is always empty because everything is stored inside location.hash. I added another useSearch hook in use-hash-location.js that retrieves query string from location.hash.

Here is a reproduction of the problem: https://codesandbox.io/p/sandbox/wouter-hash-location-t6gy4n

You can uncomment the line "{/* \<Router hook={useHashLocation2} searchHook={useSearch2}> */}" to see how the new solution works.

### DIFF
--- a/packages/wouter/src/use-hash-location.js
+++ b/packages/wouter/src/use-hash-location.js
@@ -19,8 +19,39 @@ const subscribeToHashUpdates = (callback) => {
   };
 };
 
+export const useLocationProperty = (fn, ssrFn) =>
+  useSyncExternalStore(subscribeToHashUpdates, fn, ssrFn);
+
+// when useHashLocation is used, location.search is always empty
+// so we must retrieve string from the hash
+const currentSearch = () => {
+  const hash = location.hash;
+  const hashLocation = "/" + hash.replace(/^#?\/?/, "");
+
+  const questionMarkIdx = hashLocation.indexOf("?");
+  if (questionMarkIdx !== -1) {
+    return hashLocation.slice(questionMarkIdx + 1, hashLocation.length);
+  }
+
+  return "";
+};
+
+export const useSearch = ({ ssrSearch = "" } = {}) =>
+  useLocationProperty(currentSearch, () => ssrSearch);
+
 // leading '#' is ignored, leading '/' is optional
-const currentHashLocation = () => "/" + location.hash.replace(/^#?\/?/, "");
+const currentHashLocation = () => {
+  const hash = location.hash;
+  const hashLocation = "/" + hash.replace(/^#?\/?/, "");
+
+  // remove query string
+  const questionMarkIdx = hashLocation.indexOf("?");
+  if (questionMarkIdx !== -1) {
+    return hashLocation.slice(0, questionMarkIdx);
+  }
+
+  return hashLocation;
+};
 
 export const navigate = (to, { state = null } = {}) => {
   // calling `replaceState` allows us to set the history
--- a/packages/wouter/src/use-hash-location.js
+++ b/packages/wouter/src/use-hash-location.js
@@ -19,52 +19,31 @@ const subscribeToHashUpdates = (callback) => {
   };
 };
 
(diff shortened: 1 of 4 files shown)


molefrog:
Hi, thank you for raising this PR.

As I understand, `/?abc#/` and `/#/?abc` are two different cases. The first one is the native browser search accessible via `location.search` and the second one is just a hash string that contains a `?` character. So I was wondering if we want to support this use case, or rather follow the native way of getting the search string. 

So I was thinking, maybe we can leave the hash parsing as is and instead update the `navigate` method to support strings with `?` in it (for example, calling `navigate("/abc?def")` will do two things: set the current hash and also update the search string)?

GravityTwoG:
Yep, your approach introduces fewer changes and follows the URI spec, unlike react-router's approach.

```js
export const navigate = (to, { state = null } = {}) => {
  // calling `replaceState` allows us to set the history
  // state without creating an extra entry

  let hash = to.replace(/^#?\/?/, "");
  let search = location.search;

  const searchIdx = hash.indexOf("?");
  if (searchIdx !== -1) {
    search = hash.slice(searchIdx, hash.length);
    hash = hash.slice(0, searchIdx);
  }

  history.replaceState(
    state,
    "",
    // keep the current pathname, but replace query string and hash
    location.pathname +
      search +
      // update location hash, this will cause `hashchange` event to fire
      // normalise the value before updating, so it's always preceeded with "#/"
      (location.hash = `#/${hash}`)
  );
};
```

But it will change behavior that is tested in this case:

```js
it("does not change anything besides the hash", () => {
  history.replaceState(null, "", "/foo?bar#/app");

  const { result } = renderHook(() => useHashLocation());
  const [, navigate] = result.current;

  navigate("/settings/general");
  expect(location.pathname).toBe("/foo");
  expect(location.search).toBe("?bar");
});
```

This test case actually passes, but on a navigate("/abc?def") call, it will lead to a change of location.search. Is this okay?

molefrog:
I think that is completely ok. Perhaps we can rename the test case to something like "... when doesn't contain ? symbol"

packages/wouter/src/use-hash-location.js
```diff
@@ -25,15 +25,25 @@ const currentHashLocation = () => "/" + location.hash.replace(/^#?\/?/, "");
...
   // state without creating an extra entry
+
+  let hash = to.replace(/^#?\/?/, "");
+  let search = location.search;
+
+  const searchIdx = hash.indexOf("?");
+  if (searchIdx !== -1) {
+    search = hash.slice(searchIdx, hash.length);
+    hash = hash.slice(0, searchIdx);
+  }
```
molefrog: I think this could be simplified to save up some bytes (haven't tested though)

```suggestion
  const [hash, search = location.search] = to.replace(/^#?\/?/, "").split("?")
```

packages/wouter/src/use-hash-location.js
```diff
@@ -25,22 +25,14 @@ const currentHashLocation = () => "/" + location.hash.replace(/^#?\/?/, "");
...
   history.replaceState(
     state,
     "",
     // keep the current pathname, but replace query string and hash
     location.pathname +
-      search +
+      (search ? `?${search}` : location.search) +
       // update location hash, this will cause `hashchange` event to fire
       // normalise the value before updating, so it's always preceeded with "#/"
       (location.hash = `#/${hash}`)
```
GravityTwoG, line 38: Slightly updated to preserve the `?` symbol

molefrog (approved):
Looks great!

molefrog:
Released the new version. Thank you @GravityTwoG

molefrog:
Decided to make a patch release instead, since this was a behaviour that was more or less expected.

PR #479, opened 2024-08-17
Fix #478 - fixed useParams not updating issue

### LINKED ISSUE #478: useParams do not update when there are no path params
Hello Everyone,
I ran into a small issue with the useParams() hook. I wanted to do the following in one of my projects.

1) /products/categories/:food     ->     Display Products component with food subsection
2) /products                                 ->     Display the Products component without any subsections

Both the routes are pointing to the same component. Here's the issue. 
Step 1: I'm navigating to the 1st URL (/products/categories/apple). useParams hook returns the following object -> `{"0":"apple","category":"apple"}`. 

Step 2: I am navigating to the 2nd URL (/products). Here useParams hook is pointing to the same object `{"0":"apple","category":"apple"}` and not `{}`.

I have created a codesandbox which illustrates this issue.

[Codesandbox Link](https://codesandbox.io/p/sandbox/quirky-breeze-l3lk3s)

There are three links. Click the links one by one and see how the params update. Let me know if this is the expected behavior.

khavin:
useCachedParams hook do not update the Params0 (useRef reference) value, if the previous value is a non empty object and the current value is an empty object.

More information regarding the issue: [https://github.com/molefrog/wouter/issues/478](https://github.com/molefrog/wouter/issues/478)

Updated useCachedParams to fix this. Added a test case which tests for this behavior.

### DIFF
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -198,6 +198,7 @@ const useCachedParams = (value) => {
     curr = prev.current;
 
   for (const k in value) if (value[k] !== curr[k]) curr = value;
+  if (Object.keys(value).length === 0) curr = value;
   return (prev.current = curr);
 };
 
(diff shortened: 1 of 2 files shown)


molefrog:
@khavin Thank you for spotting this bug! It looks good, I'm going to release this right away.

PR #485, opened 2024-08-28
Fix wouter link to prevent crash without href/to

dswbx:
`<Link />` components are mainly used together with `href` and/or `to` attributes, but there are cases where you'd only rely on the `onClick` property to navigate. In these cases, wouter crashes when it tries to check for absolute paths. 

This PR fixes it by making sure the absolute path test is pointed against a string and not undefined. 

---

(Context: I noticed this issue when using wouter together with [plasmic](https://www.plasmic.app/). Plasmic [allows to specify the link component](https://github.com/plasmicapp/plasmic/blob/master/packages/loader-react/src/PlasmicRootProvider.tsx#L121) being used to take advantage of client-side navigation. In some cases though you'd use the interactions (which is triggered inside `onClick`) to navigate in response to e.g. an API call. In these cases, wouter crashed).

### DIFF
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -230,7 +230,7 @@ export const Link = forwardRef((props, ref) => {
   const [currentPath, navigate] = useLocationFromRouter(router);
 
   const {
-    to,
+    to = "",
     href: targetPath = to,
     onClick: _onClick,
     asChild,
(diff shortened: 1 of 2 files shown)


packages/wouter/src/index.js
```diff
@@ -265,7 +265,9 @@ export const Link = forwardRef((props, ref) => {
 
   // handle nested routers and absolute paths
   const href = router.hrefs(
-    targetPath[0] === "~" ? targetPath.slice(1) : router.base + targetPath,
+    (targetPath ?? "")[0] === "~"
```
molefrog: Thank you for the PR!
As a suggestion, this could also be done by providing the default value for `to` when destructuring props:

```js
  const {
    to = "",
    href: targetPath = to,
```
(above in this file)

dswbx: Yes that's better, thanks! Will change it to your suggestion

molefrog (approved):

molefrog:
Released in v3.3.5

PR #491, opened 2024-10-23
feat: support server redirects

### LINKED ISSUE #468: Use wouter in SSR mode and detect redirects
Hi community, 
my question is pretty straight forward. How do I migrate from the old static-location to the new SSR mode way of working when I need to detect redirects. I think I understand the migration but im just missing the 'record: true' part and where i can make it fit into the context.

Example: I used the old functionality in this way:

`    const location = staticLocationHook(req.originalUrl, { record: true });`

And then i use it like 

```
    <Router matcher={matcher} hook={location}>
        <App />
    </Router>`
```

and also 

```
    const currentPath = location.history.slice(-1)[0];
    const redirected = currentPath !== req.originalUrl;

    if (redirected) {
      return { redirectTo: currentPath };
    }
```

Thank you in advanced

gbettencourt:
Add the ability to redirect during SSR.  Currently wouter only supports redirect once the app hydrates and the components render client side.  Being able to redirect from the server is more efficient (only requires sending a status code/path back), resulting in a better user experience.  The response can also be cached by the browser.

Resolves https://github.com/molefrog/wouter/issues/468

### DIFF
--- a/README.md
+++ b/README.md
@@ -818,13 +818,20 @@ import { Router } from "wouter";
 
 const handleRequest = (req, res) => {
   // top-level Router is mandatory in SSR mode
+  // pass an optional context object to handle redirects on the server
+  const ssrContext = {};
   const prerendered = renderToString(
-    <Router ssrPath={req.path} ssrSearch={req.search}>
+    <Router ssrPath={req.path} ssrSearch={req.search} ssrContext={ssrContext}>
       <App />
     </Router>
   );
 
-  // respond with prerendered html
+  if (ssrContext.redirectTo) {
+    // encountered redirect
+    res.redirect(ssrContext.redirectTo);
+  } else {
+    // respond with prerendered html
+  }
 };
 ```
 
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -319,9 +319,10 @@ export const Switch = ({ children, location }) => {
 
 export const Redirect = (props) => {
   const { to, href = to } = props;
-  const [, navigate] = useLocation();
+  const router = useRouter();
+  const [, navigate] = useLocationFromRouter(router);
   const redirect = useEvent(() => navigate(to || href, props));
-  const { ssrContext } = useRouter();
+  const { ssrContext } = router;
 
   // redirect is guaranteed to be stable since it is returned from useEvent
   useIsomorphicLayoutEffect(() => {
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -35,6 +35,8 @@ const defaultRouter = {
   // this option is used to override the current location during SSR
   ssrPath: undefined,
   ssrSearch: undefined,
+  // optional context to track render state during SSR
+  ssrContext: undefined,
   // customizes how `href` props are transformed for <Link />
   hrefs: (x) => x,
 };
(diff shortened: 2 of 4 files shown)


radfahrer:
@molefrog, we would greatly appreciate it if you could review this PR. We would very much like to migrate our application from `react-router` to `wouter` and this piece of missing functionality is holding us up.

README.md
```diff
@@ -818,13 +818,20 @@ import { Router } from "wouter";
 
 const handleRequest = (req, res) => {
   // top-level Router is mandatory in SSR mode
+  // pass an optional context object to handle redirects on the server
+  const ssrContext = {};
   const prerendered = renderToString(
-    <Router ssrPath={req.path} ssrSearch={req.search}>
+    <Router ssrPath={req.path} ssrSearch={req.search} ssrContext={ssrContext}>
```
molefrog, line 824: How about a shorter name `ssrCtx`?

packages/wouter/src/index.js
```diff
@@ -319,11 +321,16 @@ export const Redirect = (props) => {
...
+  const { ssrContext } = useRouter();
 
   // redirect is guaranteed to be stable since it is returned from useEvent
   useIsomorphicLayoutEffect(() => {
     redirect();
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+  if (ssrContext) {
+    ssrContext.redirectTo = to;
+  }
```
molefrog, line 334: We already use optional chaining in the library, so it is safe to write:

```suggestion
  ssrContext?.redirectTo = to;
```

packages/wouter/src/index.js
```diff
@@ -319,11 +321,16 @@ export const Redirect = (props) => {
   const { to, href = to } = props;
   const [, navigate] = useLocation();
   const redirect = useEvent(() => navigate(to || href, props));
+  const { ssrContext } = useRouter();
```
molefrog: There is an internal helper hook `useLocationFromRouter` that you can use combined with `useRouter()` to avoid extra context calls.

molefrog (changes requested):
Hey! I am so sorry for my long reply. I've added a few comments, could you take a look?

README.md
gbettencourt: Given the other props used in the above example (`ssrPath` and `ssrSearch`) IMO `Context` fits better then an abbreviation.  Not sure there's any benefit to being terse here.

packages/wouter/src/index.js
gbettencourt: This fails linting: "Parsing error: Optional chaining cannot appear in left-hand side".

packages/wouter/src/index.js
gbettencourt: Not sure I understand this suggestion, there's no need to use location (the result of `useLocationFromRouter`) in this instance.

gbettencourt: oh never-mind, I see.  I can optimize the existing call to `useLocation` above since I have a router ref.  I pushed the changes.

packages/wouter/src/index.js
molefrog: My bad, optional chaining doesn't work with assignments.

molefrog (approved):

molefrog:
Released in v3.5.0
Thank you!

PR #494, opened 2024-11-11
Update Parser type for optional loose arg

bensaufley:
Wouter _passes_ this second arg, and default parser accepts it and behaves differently when it's passed (e.g., it's needed for nested routes), so it seems useful to include in typing.


molefrog (approved):
LGTM, thank you!

PR #495, opened 2024-11-11
Export matchRoute

bensaufley:
Please feel free to reject this suggestion, but I encountered a scenario where my routes are dynamic enough that I was building a whole `Switch`ed `Route` structure dynamically that it would just be simpler to loop over my routes and `matchRoute` directly. Easy enough to just return a single, matching `Route` I suppose (which is actually what I'm doing when `nest` is at play, since there's more stuff going on there) but it _feels_ like a code smell to me to use `Route` _after_ determining that the route matched.

What do you think? Useful? Any specific reasoning behind not exporting it? Is it just not the "React Way"?

### DIFF
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -78,7 +78,7 @@ export const useSearch = () => {
   return unescape(stripQm(router.searchHook(router)));
 };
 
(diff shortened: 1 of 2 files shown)


bensaufley:
🤔 it looks like this is already exported. I don't know what happened here, sorry!

PR #497, opened 2024-11-14
Feat: memory location search hook

### LINKED ISSUE #447: Memory location does not include a search hook
When using memory location for testing purposes, query params returned by `useSearch` are ignored:

```jsx
import { render, screen } from '@testing-library/react';
import { Router, useSearch, useLocation } from 'wouter';
import { memoryLocation } from 'wouter/memory-location';

function App() {
    const search = useSearch();
    return (
        <div data-testid="test">
            search is {search}, location is {location}
        </div>
    );
}

test('Test router', () => {
    const loc = memoryLocation({ path: '/foo?key=value' });
    render(
        <Router hook={loc.hook}>
            <App />
        </Router>,
    );

    // This fails -- we get 'search is , location is /foo?key=value' instead
    expect(screen.getByTestId('test')).toEqual(
        'search is ?key=value, location is /foo?key=value',
    );
});
```

I assume this happens because no `searchHook` is passed to `Router` here, but `memoryLocation` does not appear to have one.

As a temp workaround, I can swap `useSearch()` with something like `useLocation()[0].split('?')[1]` or specify my own search hook with the query param built in, but it'd be nice not to.

mitulagr2:
Return `searchHook` from `memoryLocation`. This can be passed to `Router`.

fixes #447

### DIFF
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -7,10 +7,12 @@ import { useSyncExternalStore } from "./react-deps.js";
 
 export const memoryLocation = ({
   path = "/",
+  searchPath = "",
   static: staticLocation,
   record,
 } = {}) => {
-  let currentPath = path;
+  let currentPath = path + (searchPath && (path.split("?")[1] ? "&" : "?")) + searchPath;
+  let currentSearch = currentPath.split("?")[1] || "";
   const history = [currentPath];
   const emitter = mitt();
 
@@ -24,6 +26,7 @@ export const memoryLocation = ({
     }
 
     currentPath = path;
+    currentSearch = path.split("?")[1] || "";
     emitter.emit("navigate", path);
   };
 
@@ -39,15 +42,20 @@ export const memoryLocation = ({
     navigate,
   ];
 
+  const useMemoryQuery = () => [
+    useSyncExternalStore(subscribe, () => currentSearch)
+  ];
+
   function reset() {
     // clean history array with mutation to preserve link
     history.splice(0, history.length);
 
-    navigateImplementation(path);
+    navigateImplementation(path + (searchPath && (path.split("?")[1] ? "&" : "?")) + searchPath);
   }
 
   return {
     hook: useMemoryLocation,
+    searchHook: useMemoryQuery,
     navigate,
     history: record ? history : undefined,
     reset: record ? reset : undefined,
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -18,8 +18,7 @@ export const memoryLocation = ({
     initialPath += searchPath;
   }
 
-  let currentPath = initialPath;
-  let currentSearch = currentPath.split("?")[1] || "";
+  let [currentPath, currentSearch = ""] = initialPath.split("?");
   const history = [currentPath];
   const emitter = mitt();
 
@@ -32,8 +31,7 @@ export const memoryLocation = ({
       }
     }
 
-    currentPath = path;
-    currentSearch = path.split("?")[1] || "";
+    [currentPath, currentSearch = ""] = path.split("?");
     emitter.emit("navigate", path);
   };
 
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -47,9 +47,8 @@ export const memoryLocation = ({
     navigate,
   ];
 
-  const useMemoryQuery = () => [
-    useSyncExternalStore(subscribe, () => currentSearch),
-  ];
+  const useMemoryQuery = () =>
+    useSyncExternalStore(subscribe, () => currentSearch);
 
   function reset() {
     // clean history array with mutation to preserve link
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -19,7 +19,7 @@ export const memoryLocation = ({
   }
 
   let [currentPath, currentSearch = ""] = initialPath.split("?");
-  const history = [currentPath];
+  const history = [initialPath];
   const emitter = mitt();
 
   const navigateImplementation = (path, { replace = false } = {}) => {
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -11,7 +11,14 @@ export const memoryLocation = ({
   static: staticLocation,
   record,
 } = {}) => {
-  let currentPath = path + (searchPath && (path.split("?")[1] ? "&" : "?")) + searchPath;
+  let initialPath = path;
+  if (searchPath) {
+    // join with & if path contains search query, and ? otherwise
+    initialPath += path.split("?")[1] ? "&" : "?";
+    initialPath += searchPath;
+  }
+
+  let currentPath = initialPath;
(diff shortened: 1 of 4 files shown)


packages/wouter/test/memory-location.test.ts
```diff
@@ -23,6 +23,17 @@ it("should support initial path", () => {
   unmount();
 });
 
+it("should support initial path with query", () => {
+  const { searchHook } = memoryLocation({ path: "/test-case?foo=bar" });
+  // const { searchHook } = memoryLocation({ path: "/test-case", searchPath: "foo=bar" });
```
molefrog: Probably some leftover

tillprochaska:
Hi @mitulagr2, thanks for taking the time to fix this. I’m currently working on an app that makes heavy use of URL search parameters to encode/share UI state, so I’m quite interested in better testing support for the `useSearch` hook.

I’ve tested your changes and noticed a few issues:

* The location hook includes the search string in the returned value. The default browser location hook returns the path only.
* The search hook returns a single-valued array, whereas the browser location search hook returns a string.

I’ve tried to fix these issues here: https://github.com/tillprochaska/wouter/tree/memory-location-search-hook

@molefrog I’ve seen that this PR has been open for quite a while. Is there anything else I could help with to move this forward?

I saw that the recent release includes a new `useSearchParams` hook. It would probably be nice if the memory location implementation had support for that as well, but that could probably be added separately on top of these changes.

molefrog:
> I saw that the recent release includes a new `useSearchParams` hook. It would probably be nice if the memory location implementation had support for that as well, but that could probably be added separately on top of these changes.

`useSearchParams` uses `useSearch` internal so it should work automatically with any location hook.

molefrog:
Sadly, I started reviewing this PR a while a ago and I didn't press send.

packages/wouter/test/memory-location.test.ts
```diff
@@ -33,6 +44,16 @@ it('should return location hook that has initial path "/" by default', () => {
...
+it('should return search hook that has initial query "" by default', () => {
+  const { searchHook } = memoryLocation();
+
+  const { result, unmount } = renderHook(() => searchHook());
+  const [value] = result.current;
+
+  expect(value).toBe("");
+  unmount();
+});
+
```
molefrog, line 68: One test case that is missing is when `memoryLocation` accepts `searchPath` as a parameter.

packages/wouter/test/use-search.test.tsx
```diff
@@ -24,6 +25,18 @@ it("can be customized in the Router", () => {
...
+
+  const { result } = renderHook(() => useSearch(), {
+    wrapper: (props) => {
+      return <Router searchHook={searchHook}>{props.children}</Router>;
+    },
+  });
+
+  expect(result.current).toEqual("key=value");
+});
+
```
molefrog, line 39: Same here. Would be nice to test `searchPath` param as well.

packages/wouter/src/memory-location.js
```diff
@@ -7,10 +7,12 @@ import { useSyncExternalStore } from "./react-deps.js";
 
 export const memoryLocation = ({
   path = "/",
+  searchPath = "",
   static: staticLocation,
   record,
 } = {}) => {
-  let currentPath = path;
+  let currentPath = path + (searchPath && (path.split("?")[1] ? "&" : "?")) + searchPath;
+  let currentSearch = currentPath.split("?")[1] || "";
```
molefrog: Could this piece of code be simplified somehow?

molefrog (changes requested):

tillprochaska:
@molefrog Thanks for the update!

@mitulagr2 In case this PR isn’t relevant to you anymore or you don’t have time, feel free to let me know and I’m happy to pick up the PR and apply the requested changes :)

mitulagr2:
Thanks for the review! I'll push the changes by tomorrow

mitulagr2:
@molefrog @tillprochaska Could you please let me know if there's anything else that needs to be changed?

Currently, the memory location hook does not throw any error if both, explicit search parameter and path with search query are used simultaneously

```
const { hook, searchHook } = memoryLocation({ path: '/foo?key=value' });
<Router hook={hook} searchHook={searchHook}>

// or more explicit
const { hook, searchHook } = memoryLocation({ path: '/foo', searchPath: 'key=value' });

// also valid
const { hook, searchHook } = memoryLocation({ path: '/foo?foo=bar', searchPath: 'key=value' });
```

packages/wouter/src/memory-location.js
```diff
@@ -7,10 +7,19 @@ import { useSyncExternalStore } from "./react-deps.js";
...
 } = {}) => {
-  let currentPath = path;
+  let initialPath = path;
+  if (searchPath) {
+    // join with & if path contains search query, and ? otherwise
+    initialPath += path.split("?")[1] ? "&" : "?";
+    initialPath += searchPath;
+  }
+
+  let currentPath = initialPath;
```
tillprochaska: The default browser location implementation returns only the path without the search query:

```js
// Current browser location: /foo?key=value
const location = useLocation() // returns just /foo
```

But the current implementation includes the search query:

```js
const { hook } = memoryLocation({ path: '/foo?key=value' });
const location = useLocation() // returns /foo?key=value
```

So you’d need to do something like this here and a little further down in `navigateImplementation`:

```
let currentPath = initialPath.split("?")[0];
```

packages/wouter/src/memory-location.js
```diff
@@ -39,15 +49,20 @@ export const memoryLocation = ({
     navigate,
   ];
 
+  const useMemoryQuery = () => [
+    useSyncExternalStore(subscribe, () => currentSearch),
+  ];
```
tillprochaska: The default browser location hook returns just the query string (instead of the single-value array returned here).

tillprochaska (commented):
Hi @mitulagr2, that’s great, thank you!

I think there are still two small issues where there memory location implementation would be different from the default browser location implementation (see comments below).

mitulagr2:
Thank you for the detail! I've made said changes.

molefrog (approved):
LGTM!

packages/wouter/src/memory-location.js
```diff
@@ -7,10 +7,18 @@ import { useSyncExternalStore } from "./react-deps.js";
...
-  let currentPath = path;
+  let initialPath = path;
+  if (searchPath) {
+    // join with & if path contains search query, and ? otherwise
+    initialPath += path.split("?")[1] ? "&" : "?";
+    initialPath += searchPath;
+  }
+
+  let [currentPath, currentSearch = ""] = initialPath.split("?");
   const history = [currentPath];
```
tillprochaska: Just one more thing to fix the initial entry in the history array, now that `currentPath` doesn’t include the query string anymore.

```suggestion
  const history = [initialPath];
```

Otherwise this wouldn’t work:

```js
const { history } = memoryLocation({ path: "/foo?key=value" });
// history[0] is `/foo`, but should be `/foo?key=value`
```

tillprochaska (approved):

molefrog:
Available in the latest release!

tillprochaska:
Thank you both, much appreciated!

PR #500, opened 2024-12-20
fix(readme): typo in function name

kruschid:
The readme section above the updated code fragment says that location changes are applied by calling `navigate`. However, in the code, the function was named `setLocation`. This PR renames `setLocation` to `navigate` in the example for better clarity.
(Generally, I would prefer `navigate` over `setLocation`. `setLocation` could be misunderstood as a state setter, whereas `navigate` is more specific by pointing out the side effect. This is just an opinion. :-))


molefrog:
Thanks, good suggestion!

Issue #522, opened 2025-04-01
`option.replace` doesn't work in hash router


Aliath:
I noticed that the `navigate` function in the `useHashLocation` hook doesn’t support the `replace` option. While it does call `history.replaceState`, it also does `location.hash =` which ends up creating a new entry in the history stack instead of replacing the current one:

https://github.com/molefrog/wouter/blob/e7cb5d6072e9e184f121b9fee9463cf3d093ff64/packages/wouter/src/use-hash-location.js#L25-L40

Is there any particular reason for this behavior? Just wanted to check if this was intentional or if it might be something worth adjusting.

molefrog:
Hi! According to this comment:

```js
// update location hash, this will cause `hashchange` event to fire
```

it's needed for the event to fire, otherwise other components won't be notified. I guess I added this to be compatible with external libraries that subscribe to `hashchange` specifically (microfrontends?).

Aliath:
hmm, I'm wondering if just dispatching the `hashchange` event would be enough, instead of adding a new entry to the browser history when `option.replace` is enabled

I could put together a PR for this, but not sure if it’s worth it since the navigation hook is basically pretty small 😄

molefrog:
Closed via #523 
Released in v3.7.0

PR #523, opened 2025-04-02
feat: support `options.replace` in hash router

Aliath:
This update adds support for path replacement in the hash router using options.replace, resolving the issue referenced in https://github.com/molefrog/wouter/issues/522

### DIFF
--- a/packages/wouter/src/use-hash-location.js
+++ b/packages/wouter/src/use-hash-location.js
@@ -27,8 +27,8 @@ export const navigate = (to, { state = null, replace = false } = {}) => {
 
   const newRelativePath =
     location.pathname + (search ? `?${search}` : location.search) + `#/${hash}`;
-  const oldURL = new URL(window.location.href);
-  const newURL = new URL(newRelativePath, window.location.origin);
+  const oldURL = window.location.href;
+  const newURL = new URL(newRelativePath, window.location.origin).href;
 
   if (replace) {
     history.replaceState(state, "", newRelativePath);
--- a/packages/wouter/src/use-hash-location.js
+++ b/packages/wouter/src/use-hash-location.js
@@ -23,40 +23,25 @@ const subscribeToHashUpdates = (callback) => {
 const currentHashLocation = () => "/" + location.hash.replace(/^#?\/?/, "");
 
 export const navigate = (to, { state = null, replace = false } = {}) => {
-  // calling `replaceState` allows us to set the history
-  // state without creating an extra entry
   const [hash, search] = to.replace(/^#?\/?/, "").split("?");
 
+  const newRelativePath =
+    location.pathname + (search ? `?${search}` : location.search) + `#/${hash}`;
+  const oldURL = new URL(window.location.href);
+  const newURL = new URL(newRelativePath, window.location.origin);
+
   if (replace) {
-    history.replaceState(
-      state,
-      "",
-      // keep the current pathname, but replace query string and hash
-      location.pathname +
-        (search ? `?${search}` : location.search) +
-        `#/${hash}`
-    );
-
-    const event =
-      typeof HashChangeEvent !== "undefined"
-        ? new HashChangeEvent("hashchange")
-        : new Event("hashchange");
-
-    dispatchEvent(event);
-
-    return;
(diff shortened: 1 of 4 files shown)


packages/wouter/types/use-hash-location.d.ts
```diff
@@ -1,6 +1,9 @@
 import { Path } from "./location-hook.js";
 
-export function navigate<S = any>(to: Path, options?: { state: S }): void;
+export function navigate<S = any>(
```
molefrog, line 3: Could you also update types in wouter-preact please?

packages/wouter/src/use-hash-location.js
```diff
@@ -22,11 +22,31 @@ const subscribeToHashUpdates = (callback) => {
...
+      // keep the current pathname, but replace query string and hash
+      location.pathname +
+        (search ? `?${search}` : location.search) +
+        `#/${hash}`
+    );
+
+    const event =
+      typeof HashChangeEvent !== "undefined"
+        ? new HashChangeEvent("hashchange")
+        : new Event("hashchange");
```
molefrog: I can see that this event has also `newURL` and `oldURL` properties, maybe we should add them to the options?

packages/wouter/src/use-hash-location.js
```diff
@@ -22,11 +22,31 @@ const subscribeToHashUpdates = (callback) => {
...
 const currentHashLocation = () => "/" + location.hash.replace(/^#?\/?/, "");
 
-export const navigate = (to, { state = null } = {}) => {
+export const navigate = (to, { state = null, replace = false } = {}) => {
   // calling `replaceState` allows us to set the history
   // state without creating an extra entry
   const [hash, search] = to.replace(/^#?\/?/, "").split("?");
 
+  if (replace) {
+    history.replaceState(
```
molefrog: Not sure but, maybe we can always fire this event manually and that way we won't need to do the assigment and can simplify the code?

molefrog (changes requested):

packages/wouter/src/use-hash-location.js
Aliath: yeah that's a good call 👍

packages/wouter/src/use-hash-location.js
```diff
@@ -22,21 +22,26 @@ const subscribeToHashUpdates = (callback) => {
...
-    location.pathname +
-      (search ? `?${search}` : location.search) +
-      // update location hash, this will cause `hashchange` event to fire
-      // normalise the value before updating, so it's always preceeded with "#/"
-      (location.hash = `#/${hash}`)
-  );
+  const newRelativePath =
+    location.pathname + (search ? `?${search}` : location.search) + `#/${hash}`;
+  const oldURL = new URL(window.location.href);
+  const newURL = new URL(newRelativePath, window.location.origin);
```
molefrog: It seems that `oldURL` and `newURL` are actually strings according to the [spec](https://www.w3.org/TR/2011/WD-html5-author-20110809/history.html#event-hashchange) and they contain the full URL of the webpage including the new hash and everything else.


image

Aliath: I wondered why TS hadn't figure it out, but forgot it is a JS file 😬

packages/wouter/src/use-hash-location.js
```diff
@@ -22,21 +22,26 @@ const subscribeToHashUpdates = (callback) => {
...
-    location.pathname +
-      (search ? `?${search}` : location.search) +
-      // update location hash, this will cause `hashchange` event to fire
-      // normalise the value before updating, so it's always preceeded with "#/"
-      (location.hash = `#/${hash}`)
-  );
+  const newRelativePath =
+    location.pathname + (search ? `?${search}` : location.search) + `#/${hash}`;
+  const oldURL = window.location.href;
+  const newURL = new URL(newRelativePath, window.location.origin).href;
```
molefrog: Just to make it more consistent with the rest of the code

```suggestion
  const oldURL = location.href;
  const newURL = new URL(newRelativePath, location.origin).href;
```

molefrog (approved):
Awesome!

Issue #533, opened 2025-06-10
When location.protocol is "data:", navigation do not work.


5unnyWind:
version: v3.7.0

If `location.href` is a url like "data:text/html;base64,PHNjcmlwdD4Kd2luZG93Lm9ubWVzc2FnZSA9IC..."(which is common in figma plugins development) , the value of location.origin becomes "null", causing new URL(relativePath, location.origin) to fail during construction.

This results in the error:
Uncaught TypeError: Failed to construct 'URL': Invalid base URL
at LB (<anonymous>:190:5457)
...

molefrog:
Fixed!

PR #534, opened 2025-06-10
feat: navigate support data: protocol

5unnyWind:
#533


packages/wouter/src/use-hash-location.js
```diff
@@ -23,17 +23,26 @@ const subscribeToHashUpdates = (callback) => {
...
+    newHistoryStatePath = newHash;
+    newURL = oldURL.split("#")[0] + newHash;
+  } else {
+    const [hash, search] = to.replace(/^#?\/?/, "").split("?");
+    newHistoryStatePath =
+      location.pathname +
+      (search ? `?${search}` : location.search) +
+      `#/${hash}`;
+    newURL = new URL(newHistoryStatePath, location.origin).href;
+  }
```
molefrog, line 40: Thank you for submitting this PR. I've got a theory that if we just change the URL constructor code, we will be able to fix this without handling a special case when the protocol is `:data`

Image

Something like:
```ts
    const [hash, search] = to.replace(/^#?\/?/, "").split("?");

    const newRelativePath =
      location.pathname + (search ? `?${search}` : location.search) + `#/${hash}`;
    const oldURL = location.href;

    // Works for ALL protocols including data:
    const url = new URL(location.href);
    url.hash = `/${hash}`;
    if (search) url.search = search;
    const newURL = url.href;
```

Could you please check if this is something we could use?

molefrog: Actually, nevermind. I'm going to merge this and refactor separately. Thanks!

PR #537, opened 2025-07-25
Update README.md for memoryLocation

YutaUra:
related: https://github.com/molefrog/wouter/issues/447

If you use `useSearch` without specifying `searchHook` while using `memoryLocation`, an error will occur.
Although I understand that specifying `searchHook` is unnecessary in cases where `useSearch` is not used, I believe it would be better to include `searchHook` by default in usage examples to avoid unnecessary confusion for first-time users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -918,10 +918,10 @@ import { memoryLocation } from "wouter/memory-location";
 it("renders a user page", () => {
   // `static` option makes it immutable
   // even if you call `navigate` somewhere in the app location won't change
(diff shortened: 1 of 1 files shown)


molefrog:
Hi @YutaUra, thank you for submitting this and sorry for my delayed reply. Good catch!
I will merge this right away.

I've also realised that for better DX, `searchHook` can be automatically detected based on the hook supplied. I'm planning to work on this feature.

PR #543, opened 2025-08-27
rexport dependencies

KutnerUri:
there's no way to use the default parser outside of `useRouter()`. this PR exposes it

### DIFF
--- a/packages/wouter/src/index.js
+++ b/packages/wouter/src/index.js
@@ -28,10 +28,11 @@ import { absolutePath, relativePath, sanitizeSearch } from "./paths.js";
  * via the <Router /> component.
  */
 
-const defaultRouter = {
+export const defaultParser = parsePattern;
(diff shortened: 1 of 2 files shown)


molefrog:
Hi @KutnerUri thanks for your contribution. The default parser is actually just a function from: https://github.com/lukeed/regexparam

Would it make sense to mention it in the README instead of exporting (e.g. "the default value of `parser` in Router is `parse` function from regexparam")? Just really want to keep the bundle size lean.

KutnerUri:
yes, I ended up using that, but it seems like an implicit dependency, I'd rather use the exact same tooling as wouter.

I was working on a wrapper that adds entry and exist animations, I only needed that to use matchRoute(), which was quite inconvenient.

you can see the original code here:
https://gist.github.com/KutnerUri/777f4d59f8f81f774b0ad5f3837732ac

PR #545, opened 2025-11-22
Migrate to Vitest v4

molefrog:
- Removed Rollup bundler - no longer needed since the esm output is the same as the source and we don't need to build cjs anymore
- Migrated to Vitest v4 - updated from v3 to v4, fixed act() warnings in tests
- Fixed broken coverage reporting


molefrog:
Closing in favour of #546

PR #555, opened 2025-12-14
Handle absolute paths in Link active check

vpzomtrrfrt:


molefrog:
@vpzomtrrfrt This look like a breaking change to me, would you mind providing a bit more context on the PR?

vpzomtrrfrt:
Yeah I guess it technically might be? I assumed the existing behaviour was a bug, since it made Links not be marked as active even though they were to the current path

vydimitrov:
I came across the same issue, the Link component does not set the active class if you provide absolute path. Here is simple demo reproducing the issue - https://codesandbox.io/p/sandbox/wv6gvv

PR #557, opened 2026-02-03
feat(memory-location): add state support

### LINKED ISSUE #492: Support passing down state in `memoryLocation`.
Following the testing instructions about passing down `memoryLocation(...).hook` to `Router` I couldn't find a way to pass down state to the current location. The only args `memoryLocation` receives are:
```
  path?: Path;
  static?: boolean;
  record: boolean;
```

It would be a very useful feature to let an additional prop be passed: `state`.

wotan-allfather:
This PR adds state support to memoryLocation, similar to how useBrowserLocation handles history.state. This allows passing down state in testing scenarios and in-memory routing.

Changes:
- Add optional `state` parameter to memoryLocation options for initial state
- Accept `state` option in navigate function to update current state
- Add `state` getter property to access current state value
- Add `stateHook` for subscribing to state changes in components
- Reset state to initial value when `reset()` is called

The implementation follows the pattern suggested in issue #492 by @molefrog.

Closes #492

### DIFF
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -10,6 +10,7 @@ export const memoryLocation = ({
   searchPath = "",
   static: staticLocation,
   record,
+  state: initialState = null,
 } = {}) => {
   let initialPath = path;
   if (searchPath) {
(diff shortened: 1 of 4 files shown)


molefrog:
Thank you, it looks good! One thing though, can we bring the state generic back?

molefrog:
Sorry, have to close this in favor of https://github.com/molefrog/wouter/pull/564

Thank you for your work on this PR!

PR #561, opened 2026-03-21
feat(useSearchParams): remove trailing question mark

zsugabubus:


zsugabubus:
Initial test failure seemed unrelated. I attempted to track it down and I found lot's of flaky tests. After a hundred `bun test --randomize` iterations I think I've managed to fix everything. Should I open a separate PR?

molefrog:
hi @zsugabubus thanks for the PR!
could you summarise what was the issue you fixed here and the flaky tests?

zsugabubus:
I use `useSearchParams` hook to manage some state in the search string. Given `/path?search=hello`. When user clears the search input, I skip appending the `search` param because it looks dumb (`/path?search=`), so I just do a `prev.delete("search")` within the setter. However, the resulting URL will be `/path?` (I would expect `/path`) because the question mark remains there unconditionally.

In the added test, I set no search params and it would be `https://wouter.dev/?` without the modifications.

Regarding the tests, I saw symptoms of missing unmounts (e.g. multiple links/elements found), missing `act` warnings, and the most tricky was that a `history.back()` from a previous test could affect a later `history.pushState`+`history.length` check.

zsugabubus:
No, sorry. The most tricky was that `bun test --seed` (IIRC) is affected by file modification timestamp/inode number/directory listing order (???). So even though I had the seed number from `--randomize` or without randomization I just had a failing test, I saved the test file and the failure magically disappeared on the next run and I could get a completely new, unrelated error. It kept me a while until I realized it.

packages/wouter/test/redirect.test.tsx
```diff
@@ -1,83 +1,52 @@
...
-  const { unmount } = render(<Redirect to="/users" state={testState} />);
+  render(<Redirect to="/users" state={testState} />);
 
   expect(location.pathname).toBe("/users");
   expect(history.state).toStrictEqual(testState);
-  unmount();
-});
-
-test("useLayoutEffect should return nothing", () => {
-  const { unmount } = render(
```
molefrog, line 74: Why was this test removed? It guards against a regression where a custom location hook's `navigate` returns a truthy value (here `"foo"`) — `useLayoutEffect` would then mistakenly return that value as a cleanup function. Worth keeping unless it's intentionally obsoleted.

---

zsugabubus: a) `@ts-expect-error` already means that something is not right. If this error was caught by the type system statically, why should we run it?

b) We provide a custom hook to `<Router>` in order to override default navigation. `<Redirect>` updates that lonely `useState` only without ever interacting with the global `location`. `expect` succeeds because tests are not properly isolated and always run serially, it basically checks leftover state of the previous test (`"supports history state"`).

molefrog (approved):

zsugabubus:
Why did you choose merge commit? Do you think it is a good idea to bury my 10 LOC real modification under hundreds of unrelated changes?

molefrog:
> Why did you choose merge commit?

It was the default option, didn't think about it too much to be honest. Didn't seem like a big deal.
What is your suggestion?

zsugabubus:
> What is your suggestion?

I don't think we can change it but my personal favorite is rebase/fast-forward (after I have nice commits).

Anyway, thanks for merging that's the only thing that really matters. :)

PR #564, opened 2026-05-17
feat(memory-location): add initial location state support

### LINKED ISSUE #492: Support passing down state in `memoryLocation`.

astakhovaskold:
## Summary

Adds initial state support to memoryLocation, so tests and custom in-memory routing setups can seed the current location state and read it back from the returned object.

This change keeps the API small:

  - memoryLocation({ path, state }) now accepts an initial state
  - navigate(..., { state }) updates that state
  - the current value is exposed as memory.state

The value is intentionally exposed on the memoryLocation(...) result object rather than as a destructured snapshot, so it always reflects the latest navigation state.

  ## Changes

  - added state option to `memoryLocation`
  - stored and updated current in-memory location state alongside path/search
  - made `reset()` restore the initial state as well as the initial path
  - updated TypeScript declarations for both `wouter` and `wouter-preact`
  - added runtime and type coverage for the new API

  ## Testing

  - bun test packages/wouter/test/memory-location.test.ts
  - bun test-types

  Closes #492

### DIFF
--- a/packages/wouter/src/memory-location.js
+++ b/packages/wouter/src/memory-location.js
@@ -8,10 +8,12 @@ import { useSyncExternalStore } from "./react-deps.js";
 export const memoryLocation = ({
   path = "/",
   searchPath = "",
+  state = null,
   static: staticLocation,
   record,
 } = {}) => {
   let initialPath = path;
+  const initialState = state;
   if (searchPath) {
     // join with & if path contains search query, and ? otherwise
     initialPath += path.split("?")[1] ? "&" : "?";
@@ -19,10 +21,11 @@ export const memoryLocation = ({
   }
 
   let [currentPath, currentSearch = ""] = initialPath.split("?");
+  let currentState = initialState;
   const history = [initialPath];
   const emitter = mitt();
 
-  const navigateImplementation = (path, { replace = false } = {}) => {
+  const navigateImplementation = (path, { replace = false, state } = {}) => {
     if (record) {
       if (replace) {
         history.splice(history.length - 1, 1, path);
@@ -32,6 +35,7 @@ export const memoryLocation = ({
     }
 
     [currentPath, currentSearch = ""] = path.split("?");
+    if (state !== undefined) currentState = state;
     emitter.emit("navigate", path);
   };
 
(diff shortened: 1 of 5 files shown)


packages/wouter/src/memory-location.js
```diff
@@ -8,21 +8,24 @@ import { useSyncExternalStore } from "./react-deps.js";
...
     initialPath += searchPath;
   }
 
   let [currentPath, currentSearch = ""] = initialPath.split("?");
+  let currentState = initialState;
   const history = [initialPath];
   const emitter = mitt();
 
-  const navigateImplementation = (path, { replace = false } = {}) => {
+  const navigateImplementation = (path, { replace = false, state = null } = {}) => {
```
molefrog: Defaulting `state = null` here means every `navigate("/foo")` (no state arg) silently clears the current state. Worth deciding whether that's intended vs. preserving state when the option is omitted (`if (state !== undefined) currentState = state;`) — either way, a test pinning the behavior would help.

---

packages/wouter/src/memory-location.js
```diff
@@ -56,15 +60,21 @@ export const memoryLocation = ({
   function reset() {
     // clean history array with mutation to preserve link
     history.splice(0, history.length);
```
molefrog, line 62: Nice — restoring `initialState` here is the right fix. Note `reset` is only exposed when `record: true`, so this branch isn't currently covered by a test. A quick assertion that `reset()` rolls state back to the initial value would lock it in.

---

molefrog (commented):
Two small inline notes on `navigate` and `reset`.

---

packages/wouter/src/memory-location.js
molefrog: I would opt in for not reseting the state, since this is closer to the native browser behaviour.

astakhovaskold: Great idea

So I did this:
- `navigate("/foo")` without a state option now preserves the current state instead of clearing it
- added a test that pins this behavior

packages/wouter/src/memory-location.js
astakhovaskold: I added a dedicated test that `reset()` restores the initial state as well as the initial path

molefrog:
Thanks, planning this for the next release

leonardlib:
Hello, is there a way we can use the `state` in another component using a hook or something similar to `useLocation`? @astakhovaskold @molefrog

PR #567, opened 2026-08-07
Avoid URLSearchParams.size

akheron:
Browsers gained support for it only in 2023. See [Browser compatibility on MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/size#browser_compatibility).


molefrog (approved):
LGTM!

PR #572, opened 2026-09-13
fix(router): treat /application as outside base="/app"

damiengrandi:
## Description

`Router` with `base="/app"` treated `/application` as in-base (`lication`) because `relativePath` used a string prefix (`indexOf === 0`), not a path segment.

It now matches only when the path is the base, or continues with `/` / `%2F` (`decodeURI` leaves `%2F` encoded, so `/app%2Fusers` still yields `%2Fusers`).

## To reproduce

Save as `repro.mjs` at the repo root, then run **without** this PR:

`node repro.mjs`

```js
import { relativePath } from "./packages/wouter/src/paths.js";

const cases = [
  {
    base: "/app",
    path: "/app/users",
    expected: "/users",
    note: "real child of /app — should stay in-base",
  },
  {
    base: "/app",
    path: "/application",
    expected: "~/application",
    note: "different path that only shares the /app prefix",
  },
  {
    base: "/app",
    path: "/apple",
    expected: "~/apple",
    note: "same prefix collision",
  },
  {
    base: "/MyApp",
    path: "/MyOtherApp",
    expected: "~/MyOtherApp",
    note: "already OK — not a string prefix",
  },
];

for (const { base, path, expected, note } of cases) {
  const got = relativePath(base, path);
  const ok = got === expected;

  console.log(ok ? "ok  " : "FAIL");
  console.log(`  base=${base}  path=${path}`);
  console.log(`  expected ${JSON.stringify(expected)}`);
  console.log(`  got      ${JSON.stringify(got)}`);
  console.log(`  ${note}`);
  console.log("");
}

```

2/4 fail

## To test the fix

- Run `bun test packages/wouter/test/path-normalization.test.tsx`
- Execute `node repro.mjs` again: no fail

## Screenshots

bun test after the fix:  

image


molefrog:
Thanks, the fix is correct and brings `Router base` in line with how `<Route nest>` already matches (regexparam's loose `(?=$|\/)` lookahead). Full suite, lint and size pass here.

One request before merging: drop the `%2f` branch. `%2F` is a literal slash inside a single segment, so `/app%2Fusers` is not a child of `/app`, and regexparam already treats it that way. The extra branch also costs ~30 B in a file with 69 B of headroom. This keeps the bundle at its current size:

```js
return !base ||
  (path.toLowerCase() + "/").startsWith(base.toLowerCase() + "/")
  ? path.slice(base.length) || "/"
  : "~" + path;
```

Then flip the existing expectation to `["/app", "/app%2Fusers", "~/app%2Fusers"]`, which still verifies that `%2F` is not decoded. Tests, lint and size are green with this applied.

---

molefrog:
Hey @damiengrandi thank you for your PR. sorry for the AI-generated reply, I just asked Claude to review it. And I think it's a good point and we should drop that check. Could you please update it? I'd be happy to merge this once this fix lands.

Alexey.

damiengrandi:
Hi @molefrog, it's done!

molefrog:
Published in v3.11.1

